PR #9, opened 2017-03-06
Initial commit of Rhino plug-ins for truncated hierarchical B-splines

mennodeij:
This is the initial commit of the rhino plug-ins for gismo THB splines.

I suggest a careful review before merging this. The possible points for improvement are
* hard-coded paths to the Rhino SDK   
* hard-coded paths to an installation of G+SMO 
* inclusion of existing VCXPROJ files, instead of having CMake generate them (this seems unavoidable because of the way these projects need to be configured to compile correctly for Rhino).


filiatra:
Hi,
I am merging this and will commit some cmake fixes in a second step. As long as the plugin is not enabled, there should be no problems with this initial commit.
thnx

PR #11, opened 2017-04-12
Minor updates to ExprTk

ArashPartow:
- Updated unknown symbol resolver interface to handle all types (scalar, string and vector)
- Added compile-time check for vector indexing when using constant values
- Added return statement enable/disable via parser settings
- Added exprtk_disable_return_statement macro for disabling return statements and associated exceptions at the source code level.


ArashPartow:
Looks like valgrind fails on the clang and gcc runners.  Though at least gismo compiles.

filiatra:
Thank you Arash!

PR #22, opened 2017-11-17
This would include the unittests that are working in stable

stefantakacs:
This change does add the unittests.

It does not tell travis to run the experiments.

The interaction with devel seems to be fine.


filiatra:
Hi,
I would prefer to have the unittests as a separate repository.
People who want to get the stable branch do not need the tests probably!

PR #24, opened 2017-11-20
add a new unittest which checks the assembler

stefantakacs:


filiatra:
This looks OK, should I approve it?
then the old trunk-test can be deleted

stefantakacs:
It is more like an integration test, but I think it sould be added. Pls confirm

PR #26, opened 2017-11-21
fix move issues in gsMultiPatch

stefantakacs:


stefantakacs:
can we merge this?

seems that also in windows it does not create new issues.

filiatra:
There are some rather recent compilers that do not like the "default" keyword.
Better implement the default move constructor and assignment without using the default keyword.

I think the "default" keyword can be used only under the condition
EIGEN_GNUC_AT_MOST(4,7) && !EIGEN_COMP_PGI
but I am not completely sure,

stefantakacs:
okey, then i will change that

stefantakacs:
seems that it works on 3 sites. merging

PR #28, opened 2017-11-21
cleanup examples: unified naming, other issues

stefantakacs:


filiatra:
Hi,
The current names of the examples are mentioned in the doxygen -- updating the names must also update the docs.

Also, I would propose name_example.cpp (as in fitting_example.cpp) instead of tutorialName.cpp,

stefantakacs:
The following things should be done before merging:

I. The examples  exampleSparseSolvers and exampleLinearSolvers should be rearranged as exampleDirectLinearSolvers and exampleIterativeLinearSolvers.

II. examplePoisson and examplePoisson2 should be merged.

filiatra:
How about renaming from exampleName to name_example ? or even number them, eg. 01_name_example etc, so we can refer to example by the number ?

I will look into merging examplePoisson and examplePoisson2, in fact I want to add a new example related to this,

stefantakacs:
> I will look into merging examplePoisson and examplePoisson2, in fact I want to add a new example related to this.

Then we can merge that commit also without me merging them.

> How about renaming from exampleName to name_example ? or even number them, eg. 01_name_example etc, so we can refer to example by the number ?

That would be possible. Then we could be more consequent in having all files in unittests (maybe besides unittests/gsTutorieal)  be called "_test".

If they have numbers, this implies an ordering, where we should have an idea about it. If you have an idea, you can give me the list with all the numbers. Still, it would be problematic to add new examples somewhere in the middle.

In any case, it would be required to do the renaming with a script to fix all the docs.

Not to forget it: as there are many back-and-forth changes, it would be good to merge this branch with "squash and merge".

filiatra:
It would be nice to have numbers, but we can leave this for later..
I would go with the namings _example and _test (for tests) as you propose, hopefully you can do this easily! After the renaming we can merge this and I will cope with the poisson* in a different PR.

stefantakacs:
The files names are now (not yet pushed):

adaptiveConvectionDiffusionReaction_example.cpp
bSplineSurface_example.cpp
interpolateMap_example.cpp
poissonTHB_example.cpp
adaptRefinementTHB_example.cpp
CMakeLists.txt
knotVector_example.cpp
sparseSolvers_example.cpp
assembly_example.cpp
commandLineArg_example.cpp
linearAlgebra_example.cpp
THBRefinement_example.cpp
basis_example.cpp
fitting_example.cpp
linearSolvers_example.cpp
THBSplineBasis_example.cpp
biharmonic_example.cpp
geometry_example.cpp
mpi_example.cpp
triangulatedMeshToSolid_example.cpp
bSplineBasis_example.cpp
heatEquation_example.cpp
poisson2_example.cpp
trilinos_example.cpp
bSplineCurve_example.cpp
inputOutput_example.cpp
poisson_example.cpp
view_example.cpp

Please tell me if you suggest any changes.

PR #33, opened 2017-11-23
small config fixes

filiatra:


stefantakacs:
Seemingly, this does not work at all.

PR #40, opened 2017-11-27
Gismo uptr function

codewerfer:
Design Pattern for general use of uPtrs as return value, like it's done in clone function already. Works for 0 or more (current up to 2) function parameters - like in gsBasis::boundaryBasis - and can be easy reworked to accept more.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -83,9 +83,8 @@ clone_folder: C:\projects\gismo
 
 # branches to build
 branches:
-  # stable
-  except:
-    - gh-pages
+  only:
+    - stable
 
 # Do not build on tags (GitHub only)
 skip_tags: true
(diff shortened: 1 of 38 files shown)


codewerfer:
Windows build still don't work..
Is only "my code" checked, or is the "merged" code checked? (are that compiler errors that where also when I created this branch, and fixed in stable now, or are the new with my code?)

stefantakacs:
@codewerfer The code from this branch is checked.

codewerfer:
Reopen because got a working solution now.

stefantakacs:
Is would be clearer if "GISMO_UPTR_FUNCTION_FORWARD" was renamed to "GISMO_UPTR_FUNCTION_PURE" because it is pure virtual.

Otherwise it could be mixed up which what you call "GISMO_UPTR_FUNCTION_DEC", which is (also) a (forward) declaration, cf. https://en.wikipedia.org/wiki/Forward_declaration

filiatra:
Compilation fails on linux ?
https://cdash.ricam.oeaw.ac.at/viewBuildError.php?buildid=219506

filiatra:
Oh, probably not, this was from the other branch,

PR #43, opened 2017-11-27
Gismo uptr function merge - dont merge!!!

codewerfer:
DO NOT MERGE


filiatra:
VisualC++ looks unhappy. If nothing else works just simplify the macros in the simplest way possible to have a first version working for 0 or one arguments
Also: the "void" argument seems unnecessary, you could remove it completely..

codewerfer:
void "eats up" ..., at least with GCC 6.4.0 it looks like necessary.

Variadic Macros like `__VA_ARGS__` and ... looks like to work - the are already in the CLONE_FUNCTION macros in stable.
[cppreference.com](http://en.cppreference.com/w/cpp/preprocessor/replace) says, that the are there since C++11
but [gcc.gnu.org](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html) date them back to C99
and at least [MS introduced them in Visual C++ 2005](https://msdn.microsoft.com/en-us/library/ms177415(v=vs.90).aspx)

codewerfer:
OK, VS doesn't like any of the argument count macros.

codewerfer:
https://stackoverflow.com/questions/5530505/why-does-this-variadic-argument-count-macro-fail-with-vc

> The problem seems that Visual Studio is expanding __VA_ARGS__ after passing it into the subsequent macro, while gcc expands it before passing.
> A possible solution, as suggested by Ise Wisteria already, is to force the expansion via:

and is working as you can see with gsPreprocessor_test in cdash

PR #46, opened 2017-11-30
gsModeling: move some implementations to corresponding *.hpp files

stefantakacs:


codewerfer:
Looks OK.
gsLagrangeBasis.h is also a candidate for this sort of refactor too.

Not sure how we deal with header files at all.
Sometimes we have a gsXXXX_.cpp, where header and source meet up - like gsBasis_.cpp.
Sometimes at the end of a header file we include the corresponding source - like gsLagrangeBasis.h.
And really rarely the header file is included by the source file - like in gsMultipatchFitting.cpp

PR #48, opened 2017-11-30
future fix for GISMO_UPTR_FUNCTION macros, to accept return types of …

codewerfer:
…other classes

needed at least for boundaryBasis

### DIFF
--- a/src/gsCore/gsFunctionSet.h
+++ b/src/gsCore/gsFunctionSet.h
@@ -51,14 +51,14 @@
 // Declaration prototypes. Followed by: { ";" , "= 0;" , "{ ... }" }
 #define __DECn(n, type, name, ...)    __DEC ## n(type, name, __VA_ARGS__)
 #define __DEC0(type, name, void)      private: virtual type * name##_impl() const
-#define __DEC1(type, name, t1)        private: virtual type * name##_impl(t1) const
-#define __DEC2(type, name, t1, t2)    private: virtual type * name##_impl(t1, t2) const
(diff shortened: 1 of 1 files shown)


src/gsCore/gsFunctionSet.h
```diff
@@ -51,14 +51,14 @@
...
+#define __DEC1(type, name, t1)        private: virtual type * name##_impl(t1 n1) const
+#define __DEC2(type, name, t1, t2)    private: virtual type * name##_impl(t1 n1, t2 n2) const
 
 // Definition prototypes
-#define __DEFn(n, name, ...)          __DEF ## n(name, __VA_ARGS__)
-#define __DEF0(name, void)            public:  inline uPtr name() const { return uPtr(name##_impl()); }
-#define __DEF1(name, t1)              public:  inline uPtr name(t1 n1) const { return uPtr(name##_impl(n1)); }
-#define __DEF2(name, t1, t2)          public:  inline uPtr name(t1 n1, t2 n2) const { return uPtr(name##_impl(n1, n2)); }
+#define __DEFn(n, type, name, ...)          __DEF ## n(type, name, __VA_ARGS__)
+#define __DEF0(type, name, void)            public:  inline typename type::uPtr name() const { return uPtr(name##_impl()); }
```
stefantakacs: The MS compiler does not like superfluous uses of "typename". So, for usage in templates this should be fine, but not if type is a non-templated class.

From this point of view memory::unique_ptr<type> might be better.

PR #53, opened 2017-12-07
shorten long switch statement

filiatra:


filiatra:
This is a test, the reason is that this file takes ages to compile on visual studio..

stefantakacs:
i see.

codewerfer:
Wouldn't it be the fastest way for a compiler - and maybe also at runtime - if the whole data section contains just one or two arrays (one for x, one for w)?

Like it is now, a compiler (would) creates for every single value a entry in the .data section and read from that address. By giving all values to one or two arrays, indexed by there native indexes, the compiler should put the whole think into data section like an array (.data -> blub db 0.1, 0.2, 0.3, ...) and access by base address (blub) plus offset (size times index).

filiatra:
Do you mean as I changed it with the last commit ? Or how would you do it ?

codewerfer:
No, just some (8) multdimimensional arrays, only filled the necessary stuff - rest will be filled by compiler with 0. Cost a bit more memory 2*(33x32 + 3x65 + 3x128 + 3x255 * 8byte), but should compile and run fast, also accepts real_t as data type.

From 1 to 33 called directly by the 2 indexes (add {0} for 0 that does nothing)

But << could also be fine, didn't know that.

filiatra:
OK you can try it out if you want, but for now already it is quite faster so I am going to push it in, 
if you have a faster version we could update later.

PR #56, opened 2017-12-07
getting rid of Cxx98 troubleshooting

codewerfer:


filiatra:
Please change size_t --> index_t

PR #58, opened 2017-12-08
Cmake policy change

filiatra:


stefantakacs:
seems to contain also the expertk-commit. we can try if it works if first the other pull request is merged and then this one.

PR #61, opened 2017-12-12
fix exprtk_ad_adaptor.hpp

stefantakacs:


stefantakacs:
I think that should fix the errors; at least now the number of { coincides with the number of }

PR #64, opened 2017-12-12
add gsSparseMatrix::addTo, and small comment on setZero

filiatra:


src/gsMatrix/gsSparseMatrix.h
```diff
@@ -313,7 +318,18 @@ class gsSparseMatrix : public Eigen::SparseMatrix<T,_Options,_Index>
...
+
+private:
+    
+    /*
+      The inherited setZero() destroys column-nonzeros structure and
+      can make further computations very slow.  Almost equivalent to
+      M.setZero() is:
+      M = gsSparseMatrix(M.rows(), M.cols());
+
+      void setZero();
```
stefantakacs: This is part of the comment, not a private declaration.

filiatra: This is intended, I don't want to make it private because devel breaks.
However we have to increase awareness on this somehow, and maybe later make it private.

stefantakacs: can we declare it with GISMO_DEPRECATED ?

stefantakacs: or even write

void setZero() { *this = gsSparseMatrix(this->rows(), this->cols()); }

?

filiatra: Ok I added deprecation

PR #70, opened 2017-12-17
fix argument of free functions

filiatra:
nested is alias to const ref, 
the function will not make a copy for gsSparseMatrix<T>, but will make one for anything else (e.g. rowmajor)


filiatra:
yes, EXPORT is required for non-templated functions only

PR #74, opened 2017-12-24
Change behaviour of gsFileManager::find, and use it in gsFileData

filiatra:
Now gsFileManager::find returns the full path instead of modifiying the input, or empty string if file not found.


src/gsIO/gsFileData.hpp
```diff
@@ -56,7 +51,11 @@ gsFileData<T>::gsFileData(String const & fn)
     data = new FileData; 
     data->makeRoot();
 
-    this->read(fn); 
+    m_lastPath = gsFileManager::find(fn);
+    if ( m_lastPath.empty() )
+        gsWarn<<"gsFileData: Input file Problem: "<<fn<<" not found\n";
```
stefantakacs: Shouldn't this be an error?

Problem => problem

PR #85, opened 2018-01-16
Provide move constructor/assignment for gsFunctionExpr

stefantakacs:


stefantakacs:
This fixes a compile error from the recent expression assembler commit (for GISMO_EXTRA_DEBUG)

PR #86, opened 2018-01-16
Initial support for Ansys CFX5 import/export

mmoelle1:


mmoelle1:
The import/export functionality is currently being implemented within the MOTOR branch. Once it is completed there I will move it to the stable branch. Since I have not found a way to add new FindXYZ.cmake modules in the devel and/or MOTOR branch I have added them here.

filiatra:
I think that you can add FindXX.cmake in  devel/cmake now, so it would be better to make a PR later on when the import/export are ready

mmoelle1:
Does’t work. You include subdirectories stable before setting the path.

Matthias

mmoelle1:
BTW, can you then also remove the GISMO_DEPRECATED from the Eigen file. The splitting of G+Smo into stable and the rest makes if MUCH more complicated to change things.

Matthias

filiatra:
OK for removing the GISMO_DEPRECATED!
For the CMAKE_MODULE_PATH, we should figure this out in devel, e.g change the path when we need to ..

mmoelle1:
Ok, now it works with FindCFX5 and the corresponding changes in CmakeLists.txt in devel. I also had added “With CFX5” line to stable/cmake/gsOptions. Do you want to have a similar file in devel/cmake/gsOptions?

mmoelle1:
Can you please have a look at devel/CMakeLists.txt

if(GISMO_WITH_CFX5)
  find_package(CFX5 REQUIRED)
  set (GISMO_INCLUDE_DIRS ${GISMO_INCLUDE_DIRS} ${CFX5_INCLUDES}
  CACHE INTERNAL "${PROJECT_NAME} include directories")
  set(gismo_LINKER ${gismo_LINKER} ${CFX5_LIBRARIES}
    CACHE INTERNAL "${PROJECT_NAME} extra linker objects")
  include_directories(SYSTEM ${CFX5_INCLUDES})
  message ("  GISMO_WITH_CFX5         ${GISMO_WITH_CFX5}")
endif(GISMO_WITH_CFX5)

When this part was in stable/CMakeListsFile.txt, the extra include files were found and gsScrewMachineCFX.cpp compiled fine. Now, these include files are still found by the FindCFX5 macro but the include search path is not adjusted correctly. BTW, we must no add these files to G+Smo since they are part of the commercial product Ansys CFX5, which the user must have installed him/herself.

Matthias

PR #91, opened 2018-01-24
Unify output behavior

stefantakacs:
* Never write directly to std::cout or std::cerr
* In general, the library itself should not write to gsInfo, only to gsDebug and gsWarn
* For print commands, it should be possible to provide the stream. For bw-compat, defaults are provided.

Changes have been done for src, unittests and examples.


filiatra:
(as long as the tests pass..)

PR #93, opened 2018-01-26
Some improvements for gsDebug.h:

stefantakacs:
* Unify the wording of GISMO_ASSERT and GISMO_ENSURE
* Write errors (also from GISMO_ASSERT) to gsWarn as they are real errors.
* Use "do {...} while(false)" everywhere to obtain a unified look-and-feel.


filiatra:
look ok to me as well

PR #94, opened 2018-01-26
Updates for expressions, sparse matrix additions and other smaller issues

filiatra:


src/gsAssembler/gsExpressions.h
```diff
@@ -2844,6 +2868,7 @@ GISMO_SHORTCUT_VAR_EXPRESSION(  div, jac(u).trace() )
 GISMO_SHORTCUT_MAP_EXPRESSION(unv, nv(G).normalized()   ) //(!) bug + mem. leak
 
 GISMO_SHORTCUT_PHY_EXPRESSION(igrad, grad(u)*jac(G).ginv() ) // transpose() problem ??
+GISMO_SHORTCUT_VAR_EXPRESSION(igrad, grad(u) ) // u is presumed to be defined over G
```
stefantakacs: Does this now mean igrad == grad ?

filiatra: yes, user is aware of where the function is defined..

src/gsAssembler/gsExprHelper.h
```diff
@@ -136,7 +136,7 @@ class gsExprHelper
         m_vlist.push_back( expr::gsFeVariable<T>() );
         expr::gsFeVariable<T> & var = m_vlist.back();
         gsFuncData<T> & fd = m_ptable[&mp];
-        fd.dim = mp.dimensions();
+        //fd.dim = mp.dimensions();
```
codewerfer: Not sure if this is correct. I mean, I said that gcc 4.8.x fails at this point (or better inside dimensions(), where the linker can't find the correct implementation for targetDim()), but I didn't meant that it should be kicked out.

src/gsMatrix/gsSparseMatrix.h
```diff
@@ -319,6 +319,102 @@ class gsSparseMatrix : public Eigen::SparseMatrix<T,_Options,_Index>
...
+                 outer.begin(); k!=outer.end();++i,++j,++k)
+        {
+            *i = this->innerIndexPtr() + this->outerIndexPtr()[*k  ];
+            *j = this->innerIndexPtr() + this->outerIndexPtr()[*k+1];
+        }
+
+        const index_t idMax = this->cols();
+        index_t curMin;
+
+        while (true)
```
codewerfer: Can't see the abort condition.

src/gsAssembler/gsExpressions.h
stefantakacs: i am still not yet sure how to work with the assembler in this case.

src/gsAssembler/gsVisitorMass.h
```diff
@@ -76,9 +76,19 @@ class gsVisitorMass
                          gsGeometryEvaluator<T> & geoEval,
                          gsVector<T> const      & quWeights)
     {
+        // /* // standard version
+        for (index_t k = 0; k != quWeights.rows(); ++k) // loop over quadrature nodes
```
stefantakacs: do we know if the change has any effect on the efficiency?

src/gsAssembler/gsExprHelper.h
filiatra: I took it out for other reasons, it is not needed here..

src/gsAssembler/gsVisitorMass.h
filiatra: This was unintentional, will revert it

src/gsAssembler/gsExpressions.h
filiatra: its 2 different commands. 
igrad(f,G) is the isogeometric gradient of paremeter-defined function
igrad(f) is the isogeometric gradient of function defined on physical coordinates.
I don't know a better way to express this..

PR #95, opened 2018-01-30
fix memory leak in gsFunctionExpr.hpp

codewerfer:
tested with GCC 6.4.0 Cxx 98 and 11


stefantakacs:
@codewerfer If you think that the new version is fine, pls merge.

codewerfer:
@stefantakacs THX :)

PR #96, opened 2018-02-02
Remove ptrs

codewerfer:
* remove of gsMultiPatch pointers where possible (and useful)
Still needs to be done also in trunk.


plugins/gsAxel/gsGeometryCreator.cpp
```diff
@@ -216,8 +216,8 @@ void gsGeometryCreator::loadMultiPatch(void)
 
     if ( fileName.size() )
     {
-        gismo::gsMultiPatch<double>::uPtr myGismoData = gismo::gsReadFile<double>(fileName.toUtf8().constData() ) ;
-        if ( myGismoData)
+        gismo::gsMultiPatch<double> myGismoData = gismo::gsReadFile<double>(fileName.toUtf8().constData() ) ;
```
codewerfer: Need somebody with Axel to verify this.

src/gsAssembler/gsAssembler.hpp
```diff
@@ -623,9 +623,9 @@ template<class T>
 gsField<T> gsAssembler<T>::constructSolution(const gsMatrix<T>& solVector,
                                                 int unk) const
 {
-    gsMultiPatch<T> * result = new gsMultiPatch<T>;
+    typename gsMultiPatch<T>::Ptr result(new gsMultiPatch<T>);
```
codewerfer: Makes for me more sense. Pointer needed to survive scope, therefor I use a smartpointer.

src/gsAssembler/gsExprAssembler.h
```diff
@@ -1221,7 +1221,7 @@ void gsExprAssembler<T>::computeDirichletDofsL2Proj(const expr::gsFeSpace<T>& u)
     gsMatrix<unsigned> globIdxAct;
     gsMatrix<T> basisVals;
 
-    const gsMultiPatch<T> * mp = (const gsMultiPatch<T>*)(& m_exprdata->getMap().source());
+    const gsMultiPatch<T> mp = *(const gsMultiPatch<T> *)(& m_exprdata->getMap().source());
```
codewerfer: Don't like this line.

src/gsCore/gsMultiPatch.h
```diff
@@ -168,6 +168,9 @@ class gsMultiPatch : public gsBoxTopology, public gsFunctionSet<T>
     /// manifold (ie. it has no boundaries)
     bool isClosed() { return this->nBoundary() == 0; }
 
+    /// \brief Returns true if gsMultiPatch is empty.
+    const bool isEmpty() const { return this->nPatches() <= 0; }
```
codewerfer: As I think, this function is necessary, at least when we read files with gsReadFile.

src/gsIO/gsReadFile.h
```diff
@@ -44,8 +44,12 @@ class gsReadFile
...
 
     /** 
-        \brief Opens a file and reads an object into a pointer. This is
-        used in conjuction with a cast operator to a pointer type
+        \brief Opens a file and reads an object into a smartpointer
+        (uPtr) or an object itself. This is used in conjuction with
+        cast operators and depends on the type of object. Most types
+        are casted to uPtr. gsMultiPatch will be a gsMultiPatch object.
+        For gsBasis there exists a cast opterator to
+        std::vector<gsBasis<T>::uPtr>.
```
codewerfer: What should be the result of a non working operation?

src/gsCore/gsMultiPatch.h
filiatra: To conform to std, please rename it to "bool empty() const", and implement as { return m_patches.empty(); }

src/gsIO/gsReadFile.h
filiatra: gsReadFile should always cast to pointer, I removed the cast to multipatch and reading as object should be done as in  examples/gsPatchFromBoundary.cpp, see
https://github.com/gismo/gismo/pull/96/commits/d97380abba6ce7dcfbc2a7fc581277a6a3fa61ca

plugins/gsAxel/gsGeometryCreator.cpp
filiatra: just change it back..

src/gsIO/gsFileData.h
```diff
@@ -344,8 +344,11 @@ class gsFileData
             gsWarn <<"gsFileData: getAnyFirst: Didn't find any "<<
                 internal::gsXml<Object>::type()<<" "<< 
                 internal::gsXml<Object>::tag() <<". Error.\n";
-      }
-        internal::gsXml<Object>::get_into(node, result);// Using gsXmlUtils
+        }
```
codewerfer: Or should we throw here a std::runtime_error through GISMO_ENSURE?

stefantakacs: No, that would cause massive changes because all the gsFileData functions only write to gsWarn and continue.

filiatra: I think it is fine as you did it..

stefantakacs: yes, this is what i tried to say.

DaHofa:
What was the reason, why the Ptr of gsMultiPatch got removed? Now the gsNurbsCreator returns unique pointers, but only the multipatches are objects. I am not sure if this makes sense.

stefantakacs:
@DaHofa  The idea behind this is to use pointers in general only for polymorphic classes.

DaHofa:
if we do not use c++11, then we make always a copy, when returning multipatches. 

Btw, I not aware of any C++ book/guidline, which says that this should be the preferred way and I find this concept quite complicated. I think then this also applies to gsMatrix and gsSparseMatrix, isnt it?

stefantakacs:
Performance on c++98/03 should not be overrated due to @filiatra. For newer compilers, everything should typically work with RVO.

To return matrices and vectors by value is at least in Gismo currently the standard. Only some functions do it differently.

filiatra:
Hi,
In c++ 11 or newer the recomended way to retun is an lvalue.
In previous c++ this was a problem, but RVO fixed it
I think that pointers in places that they should not exist make things un-necessarily complicated..

filiatra:
See for instance https://isocpp.org/blog/2013/02/no-really-moving-a-return-value-is-easy-stackoverflow

PR #99, opened 2018-02-07
fix src/gsCore/gsField.h 

stefantakacs:


src/gsCore/gsField.h
```diff
@@ -234,7 +234,7 @@ class gsField
     int dim() const { return m_fields->targetDim(); }
 
     /// Returns the number of patches.
-    int nPatches()  const { return m_patches->size(); }
+    int nPatches()  const { return dynamic_cast<const gsMultiPatch<T>*>(m_patches)->nPatches(); }
```
filiatra: dynamic_cast may fail, better add an assert as well..

filiatra:
Seems there is a problem:
https://cdash.ricam.oeaw.ac.at/viewTest.php?onlydelta&buildid=240251

better implement nPatches as nPieces, since domain can be either geometry or multipatch..

PR #100, opened 2018-02-07
fix bug on dofmapper, and a typo

filiatra:


src/gsCore/gsDofMapper.cpp
```diff
@@ -114,7 +114,10 @@ void gsDofMapper::matchDof( index_t u, index_t i, index_t v, index_t j )
     else if (d1 == 0)   // first dof is a free dof
     {
         if (d2 == 0)
+        {
```
stefantakacs: I do not understand what's going on here and why we return after adding a dof coupling.

filiatra: before, the code assumed that 2 dofs which are coupled together are 2 different dofs. 
So if I match some dof with itself then there is a segfault, because this case was not checked and not treated.

Now with this addition one can mark that a dof is "coupled" even though it is just marked as coupled and not colapsed with a different dof. This is usefull and came up because I needed to just mark a dof as coupled..

stefantakacs: thnx

PR #102, opened 2018-02-08
change assert to throw exception, change CG to use full matrix

filiatra:
CG is more efficient when using full matrix


stefantakacs:
that's at least fine for gcc.

filiatra:
thnx!

PR #104, opened 2018-02-12
Change return types, remove obosete codes and minor fixes

filiatra:
remove oboslete constructors from gsMatrix,
add handy members to gsDofMapper
change return types from pointer to gsMatrix


filiatra:
@codewerfer Please try to eliminate similar return type problems in all the code..

PR #105, opened 2018-02-13
Gs matrix

codewerfer:
* removed all safe calls from stable
* removed gsMatrix::uPtr

### DIFF
--- a/examples/gsInterpolateMap.cpp
+++ b/examples/gsInterpolateMap.cpp
@@ -113,10 +113,10 @@ int main(int argc, char* argv[])
         tBasis = gsBSplineBasis<>::make(KV);
         break;
     case 2:
-        tBasis = memory::unique_ptr<gsTensorBSplineBasis<2> >(new gsTensorBSplineBasis<2>(KV,KV));
+        tBasis = memory::make_unique(new gsTensorBSplineBasis<2>(KV,KV));
         break;
     case 3:
-        tBasis = memory::unique_ptr<gsTensorBSplineBasis<3> >(new gsTensorBSplineBasis<3>(KV,KV,KV));
+        tBasis = memory::make_unique(new gsTensorBSplineBasis<3>(KV,KV,KV));
         break;
     default:
     {
--- a/examples/gsInterpolateMap.cpp
+++ b/examples/gsInterpolateMap.cpp
@@ -113,10 +113,10 @@ int main(int argc, char* argv[])
         tBasis = gsBSplineBasis<>::make(KV);
         break;
     case 2:
-        tBasis = safe( new gsTensorBSplineBasis<2>(KV,KV) );
+        tBasis = memory::unique_ptr<gsTensorBSplineBasis<2> >(new gsTensorBSplineBasis<2>(KV,KV));
         break;
     case 3:
-        tBasis = safe( new gsTensorBSplineBasis<3>(KV,KV,KV) );
+        tBasis = memory::unique_ptr<gsTensorBSplineBasis<3> >(new gsTensorBSplineBasis<3>(KV,KV,KV));
         break;
     default:
     {
--- a/examples/poissonTHB_example.cpp
+++ b/examples/poissonTHB_example.cpp
@@ -87,7 +87,7 @@ int main(int argc, char *argv[])
   //gsFunctionExpr<>  f("if( (x-0.25)^2 + (y-1.6)^2 < 0.2^2, 1, 0 )",2);
   gsFunctionExpr<>  g("0",2);
   gsMultiPatch<> patches( *gsNurbsCreator<>::BSplineRectangle(0.0,0.0,2.0,1.0) );
-  //gsMultiPatch<> patches( *safe(gsNurbsCreator<>::BSplineFatQuarterAnnulus(1.0, 2.0)) );
+  //gsMultiPatch<> patches( gsNurbsCreator<>::BSplineFatQuarterAnnulus(1.0, 2.0) );
 
   //RefineLoopMax = 6;
   //refParameter = 0.6;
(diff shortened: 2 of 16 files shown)


examples/gsInterpolateMap.cpp
```diff
@@ -113,10 +113,10 @@ int main(int argc, char* argv[])
         tBasis = gsBSplineBasis<>::make(KV);
         break;
     case 2:
-        tBasis = safe( new gsTensorBSplineBasis<2>(KV,KV) );
+        tBasis = memory::unique_ptr<gsTensorBSplineBasis<2> >(new gsTensorBSplineBasis<2>(KV,KV));
```
stefantakacs: I don't like that. We have wrapper functions not to be required to specify the template parameter. That's even what std does.

We might change this to memory::make_unique(...) if desired.

examples/poissonTHB_example.cpp
```diff
@@ -103,8 +103,8 @@ int main(int argc, char *argv[])
   gsFunctionExpr<>  f("0",2);
   gsFunctionExpr<>  g("if( y>0, ( (x^2+y^2)^(1.0/3.0) )*sin( (2*atan2(y,x) - pi)/3.0 ), ( (x^2+y^2)^(1.0/3.0) )*sin( (2*atan2(y,x) +3*pi)/3.0 ) )",2);
 
-  //gsMultiPatch<> patches( *safe(gsNurbsCreator<>::BSplineLShape_p2C0()) );
-  gsMultiPatch<> patches( *safe(gsNurbsCreator<>::BSplineLShape_p1()) );
+  //gsMultiPatch<> patches( gsNurbsCreator<>::BSplineLShape_p2C0() );
+  gsMultiPatch<> patches( gsNurbsCreator<>::BSplineLShape_p1() );
```
stefantakacs: Does this work?

examples/poisson_example.cpp
```diff
@@ -63,7 +63,7 @@ int main(int argc, char *argv[])
     // For single patch unit square of quadratic elements use (Note:
     //you need to update the bounadry conditions section for this to
     //work properly!) :
-    //patches = safe( new gsMultiPatch<>(*gsNurbsCreator<>::BSplineSquare(2)));
+    //patches = new gsMultiPatch<>(*gsNurbsCreator<>::BSplineSquare(2));
```
stefantakacs: remove "new" here. otherwise, it does not work.

src/gsCore/gsRationalBasis.h
```diff
@@ -219,7 +219,7 @@ class gsRationalBasis : public gsBasis< typename SrcT::Scalar_t >
       for ( index_t i=0; i<ind.size(); ++i)
       ww(i,0) = m_weights( (*ind)(i,0), 0);
         
-      return new BoundaryBasisType(*safe(bb), give(ww));
+      return new BoundaryBasisType((bb), give(ww));
```
stefantakacs: (bb) => bb

src/gsIO/gsXmlUtils.hpp
```diff
@@ -415,7 +415,7 @@ class gsXml< gsNurbsBasis<T> >
         
         gsXmlNode * tmp = node->first_node("KnotVector");
         // if type: == Plain, == Compact .. 
-        gsKnotVector<T> kv = * safe( gsXml<gsKnotVector<T> >::get (tmp) );
+        gsKnotVector<T> kv = ( gsXml<gsKnotVector<T> >::get (tmp) );
```
stefantakacs: This doesn't work. You cannot cast a raw pointer to an object. Moreover, it's a memory leak.

src/gsNurbs/gsBSpline.h
```diff
@@ -403,7 +403,7 @@ gsBSpline<T> operator*(const gsBSpline<T> & lhs, const gsBSpline<T> & rhs)
     const gsMatrix<T> ev = (lhs.eval(pts)).array() * (rhs.eval(pts)).array();
     
     // fixme: avoid temporaries here
-    return *safe(static_cast<gsBSpline<T>*>(gsBSplineBasis<T>(kv).interpolateData(ev,pts)));
+    return (static_cast<gsBSpline<T>*>(gsBSplineBasis<T>(kv).interpolateData(ev,pts)));
```
stefantakacs: cannot cast gsBSpline&lt;T&gt;* to gsBSpline&lt;T&gt;. Moreover, it's a memory leak.

src/gsTensor/gsTensorBasis.hpp
```diff
@@ -313,9 +313,9 @@ gsMatrix<unsigned> gsTensorBasis<d,T>::allBoundary() const
        {
        bdofs= dofs / size(k);
        r-=  bdofs;
-       res->block(r,0, bdofs, 1) = *safe( this->slice(i, size(k) - 1) );
+       res->block(r,0, bdofs, 1) = ( this->slice(i, size(k) - 1) );
```
stefantakacs: same as above

codewerfer: It's code that is for more then 3 years out-commented.

stefantakacs: It could be removed then.

codewerfer: Ahh.. I see, it should be some type of example. Ignored yesterday every out-commended code and looked fast over it without any recognition.

filiatra: Just remove this commented part, it is too old..

src/gsIO/gsXmlUtils.hpp
filiatra: Just remove the comment, it is also to old..

src/gsModeling/gsPlanarDomain.h
```diff
@@ -21,6 +21,7 @@
 
 #include <gsModeling/gsCurveLoop.h> 
 #include <gsModeling/gsTemplate.h>
+#include <gsUtils/gsMesh/gsMesh.h>
```
filiatra: Does it work if you include gsMesh.h in gsPlanarDomain.hpp ? This would be more preferable

codewerfer: It's then needed in gsTrimSurface.h

Yesterday I build it without include and with memory::unique_ptr<gsMesh<T> > instead of any gsMesh<T>::uPtr in gsPlanar and gsTrimSurface. It took on my computer about 18s longer (r: 6m27.940s, u:17m16.519s vs. 6m35.502s, u:17m27.351s) to build.

PR #106, opened 2018-02-14
Penalty parameter updated for 1D case

roeltielen24:
In the 1D case with Nitsche's method for BC, we have element.getCellSize() = 0, leading to mu = Inf

Therefore, the following updates are proposed. 

-----

### DIFF
--- a/src/gsAssembler/gsVisitorNitsche.h
+++ b/src/gsAssembler/gsVisitorNitsche.h
@@ -134,7 +134,8 @@ public:
         geoEval.transformGradients(k, bGrads, pGrads);
         
         // Get penalty parameter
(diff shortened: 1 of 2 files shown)


filiatra:
I proposed the 2, for no good reason :)

stefantakacs:
I would have proposed 1. Because then the penalty parameter mu is equal to penalty, i.e., you get what you ask for.

filiatra:
yes, one is more conforming, oeltielen24 could change this

roeltielen24:
Value has been changed to 1 as requested

stefantakacs:
Thanks.

@roeltielen : Do you have the rights to merge it yourself?

roeltielen24:
yes :-)

PR #108, opened 2018-02-22
Permutation of dofmapper indices

DaHofa:
This should allows for more flexibility, when using the dofmapper. The user can now permute the ordering of the dofs. The only restriction is, that the eliminated dofs have to be still ordered last (makes no sense otherwise).  Moreover, it is only allowed to perform a  permutation once (see also flag m_isPermuted). This could be implemented at some point, but I think is not required now. The permutation can be combined before.

It is then getting a bit more complex with the coupled dofs, since they may be mixed with the pure free dofs. In order to have a fast checking if a dofs is coupled (and which coupled idx it is), while having a small memory consumption, additional members are added to store precomputed information. Other functions are not affected.

### DIFF
--- a/src/gsCore/gsDofMapper.cpp
+++ b/src/gsCore/gsDofMapper.cpp
@@ -182,6 +182,7 @@ void gsDofMapper::markBoundary( index_t k, const gsMatrix<unsigned> & boundaryDo
 
 void gsDofMapper::markCoupledAsTagged()
 {
+    GISMO_ASSERT(m_curElimId==0, "finalize() was not called on gsDofMapper");
     m_tagged.reserve(m_tagged.size()+m_numCpldDofs);
     for(int i=0; i< m_numCpldDofs;++i)
         m_tagged.push_back(m_numFreeDofs-m_numCpldDofs+i);
@@ -305,8 +306,8 @@ void gsDofMapper::permuteFreeDofs(const gsVector<index_t>& permutation)
     //make a copy of the old ordering, easiest way to implement the permutation. Inplace reordering is quite hard.
     std::vector<index_t> dofs = m_dofs;
     //first use a tempory coupled vector, other wise is_coupled_index() would not work correctly
-    std::vector<index_t> tagged;
-    tagged.reserve(taggedSize()); //reserve enough memory
+    std::vector<index_t> tagged_permuted;
+    tagged_permuted.reserve(taggedSize()); //reserve enough memory
     
     for(index_t i=0; i<(index_t)dofs.size();++i)
     {
@@ -316,10 +317,12 @@ void gsDofMapper::permuteFreeDofs(const gsVector<index_t>& permutation)
             m_dofs[i] = permutation[idx];
             //fill  bookkeeping for tagged dofs
             if(is_tagged_index(idx))
-                tagged.push_back(m_dofs[i]);
+                tagged_permuted.push_back(m_dofs[i]);
         }
+        else if(is_tag_index(idx)) //Take care about eliminated tagged dofs
+	    tagged.push_back(idx);
     }
-    m_tagged.swap(tagged);
+    m_tagged.swap(tagged_permuted);
 
     //sort and delete the duplicated ones
     std::sort(m_tagged.begin(),m_tagged.end());
--- a/src/gsCore/gsDofMapper.cpp
+++ b/src/gsCore/gsDofMapper.cpp
@@ -18,7 +18,7 @@ namespace gismo
 {
 
 gsDofMapper::gsDofMapper() : 
-m_shift(0), m_numFreeDofs(0), m_numCpldDofs(1), m_curElimId(-1),m_isPermuted(false)
+m_shift(0), m_numFreeDofs(0), m_numCpldDofs(1), m_curElimId(-1)
 { 
     m_offset.resize(1,0);
 }
@@ -256,7 +256,6 @@ void gsDofMapper::finalize()
     m_numElimDofs = curElimDof - m_numFreeDofs;
 
     m_curElimId = 0;// Only equal to zero after finalize is called.
-    m_isPermuted = false;
 }
 
 std::ostream& gsDofMapper::print( std::ostream& os ) const
@@ -287,16 +286,13 @@ void gsDofMapper::permuteFreeDofs(const gsVector<index_t>& permutation)
 {
     GISMO_ASSERT(m_curElimId==0, "finalize() was not called on gsDofMapper");
     GISMO_ASSERT(m_numFreeDofs == permutation.size(), "permutation size does not match number of free dofs");
-    GISMO_ASSERT(m_isPermuted==false, "you cannot permute the dofVector twice, combine the permutation");
-
+    GISMO_ASSERT(m_coupled.isEmpty(), "you cannot permute the dofVector twice, combine the permutation");
+    GISMO_ASSERT(std::is_permutation(permutation.begin(), permutation.end()), "input vector is not a permutation");
 
     //make a copy of the old ordering, easiest way to implement the permutation. Inplace reordering is quite hard.
     std::vector<index_t> dofs = m_dofs;
-    m_isCoupledIdx.clear();
-    m_isCoupledIdx.resize(m_numFreeDofs,false);
-    m_coupledIndexNumber.clear();
-
-    index_t cIdx = 0;
+    m_coupled.reserve(2*coupledSize()); //reserve enough memory
+    
     for(index_t i=0; i<(index_t)dofs.size();++i)
     {
         index_t idx = dofs[i];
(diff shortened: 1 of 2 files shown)


src/gsCore/gsDofMapper.cpp
```diff
@@ -21,6 +21,8 @@ gsDofMapper::gsDofMapper() :
 m_shift(0), m_numFreeDofs(0), m_numCpldDofs(1), m_curElimId(-1)
 { 
     m_offset.resize(1,0);
+
+    m_isPermuted = false;
```
stefantakacs: maybe we should set m_isPermuted in the initializer list; that would make it easier to check that it is initialized in all cases.

src/gsCore/gsDofMapper.h
```diff
@@ -443,6 +459,13 @@ class GISMO_EXPORT gsDofMapper
     // After finalize() is called m_curElimId takes the value zero.
     index_t m_curElimId;
 
+    //stores if the dofmapper got reordered
+    bool m_isPermuted;
+    //vector of length m_numFreeDofs, which stores if a dof is also coupled. (only used after permutation)
+    std::vector<bool> m_isCoupledIdx;
+    //maps a coupled index to its corresponding number, i.e. 22 -> 2, the free dof 22 is the third coupled index (only used after permutation)
+    std::map<index_t,index_t> m_coupledIdxCount;
```
stefantakacs: Why count?

Shouldn't it be index?

src/gsCore/gsDofMapper.cpp
DaHofa: jup, you are right

src/gsCore/gsDofMapper.h
DaHofa: m_coupledIndexNumber?

filiatra:
@DaHofa: Do you ever need the isCoupled functions in your codes after you do the permutation ?

DaHofa:
yes, it is needed. This is also, why I encountered the need for changing the function, while looking for the bugs ;)

filiatra:
I find the solution a little ad hoc, I leave it open to think more on this.
I also would like for another application to add different attributes to the dofs, so I will try to think of a more general solution for this..

DaHofa:
well, it fulfils its task. Due to the mixing of coupled and free dofs, one cannot use the fast option before.  However, the coupled dofs stuff is anyway used quite rarely, so a slight performance drop in this function will not effect the overall performance. 

Anyway, is there a chance to merge this part (as a preliminary merge) and continue developing this branch further. I would need this functionality to commit the stuff in devel.

filiatra:
OK, then I can propose some changes first.
Also: do you need both functions? what is the difference in these 2? they mostly look the same ?

DaHofa:
fine, no i need only permuteFree, (I think you ask me to implement both). permute() reorders all dofs, but this one can be deleted.

src/gsCore/gsDofMapper.cpp
```diff
@@ -282,6 +283,69 @@ void gsDofMapper::setIdentity(index_t nPatches, size_t nDofs)
...
 
+void gsDofMapper::permuteFreeDofs(const gsVector<index_t>& permutation)
+{
+    GISMO_ASSERT(m_curElimId==0, "finalize() was not called on gsDofMapper");
+    GISMO_ASSERT(m_numFreeDofs == permutation.size(), "permutation size does not match number of free dofs");
+    GISMO_ASSERT(m_isPermuted==false, "you cannot permute the dofVector twice, combine the permutation");
+
+
+    //make a copy of the old ordering, easiest way to implement the permutation. Inplace reordering is quite hard.
+    std::vector<index_t> dofs = m_dofs;
```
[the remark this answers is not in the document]
DaHofa: I though about this again, but wont work. consider the vector {1,2,3,4} and the permutation {4,3,2,1}. Then you write at the first place 4, at the second 3, but at the third again 3 and the last again 4, since you override the entries.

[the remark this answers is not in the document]
filiatra: Yes, you are right, now I see it, please ignore the comment

src/gsCore/gsDofMapper.cpp
```diff
@@ -282,6 +283,69 @@ void gsDofMapper::setIdentity(index_t nPatches, size_t nDofs)
     m_dofs.resize( m_numFreeDofs, 0);
 }
 
+void gsDofMapper::permuteFreeDofs(const gsVector<index_t>& permutation)
+{
```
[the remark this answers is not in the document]
DaHofa: std::is_permutation requires c++11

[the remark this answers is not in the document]
stefantakacs: might be
\#if __cplusplus > ....
GISMO_ASSERT(...)
\#endif

DaHofa:
so, the proposed changes are included. Next, I include the tagged dofs and change the Dofmapper as discussed with @filiatra

DaHofa:
@filiatra please have a look at the tagged version of DofMapper, my code still works :)

DaHofa:
More Documentation will be added on Monday

filiatra:
OK it looks good, probably I can also use it now.
I will add some more comments though for Monday :)

src/gsCore/gsDofMapper.cpp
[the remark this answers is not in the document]
filiatra: @stefantakacs seems that this function cannot be used here (it checks two different vectors for being permutation of each other...)

src/gsCore/gsDofMapper.cpp
```diff
@@ -175,6 +180,13 @@ void gsDofMapper::markBoundary( index_t k, const gsMatrix<unsigned> & boundaryDo
     // TO DO: save inverse, eg boundaryDofs(i,0)
 }
 
+void gsDofMapper::markCoupledAsTagged()
+{
+    m_tagged.reserve(m_tagged.size()+m_numCpldDofs);
+    for(int i=0; i< m_numCpldDofs;++i)
+        m_tagged.push_back(m_numFreeDofs-m_numCpldDofs+i);
+}
+
```
filiatra: This function seems buggy: it should insert into m_tagged some indices from m_dofs, not the "cindex" !

src/gsCore/gsDofMapper.cpp
```diff
@@ -282,6 +295,42 @@ void gsDofMapper::setIdentity(index_t nPatches, size_t nDofs)
...
+    GISMO_ASSERT(m_curElimId==0, "finalize() was not called on gsDofMapper");
+    GISMO_ASSERT(m_numFreeDofs == permutation.size(), "permutation size does not match number of free dofs");
+    //GISMO_ASSERT(m_tagged.empty(), "you cannot permute the dofVector twice, combine the permutation");
+    //We could check here, that permutation is indeed a permutation
+
+    //make a copy of the old ordering, easiest way to implement the permutation. Inplace reordering is quite hard.
+    std::vector<index_t> dofs = m_dofs;
+    //first use a tempory coupled vector, other wise is_coupled_index() would not work correctly
+    std::vector<index_t> tagged;
+    tagged.reserve(taggedSize()); //reserve enough memory
```
filiatra: Would be better if tagged where not changed inside this function..
You could do in your codes:

mapper.markCoupledAsTagged();
mapper.permuteFreeDofs(..)

and that is all

src/gsCore/gsDofMapper.h
```diff
@@ -314,6 +324,16 @@ class GISMO_EXPORT gsDofMapper
...
     }
 
+    /// @brief Returns the tagged dof index
+    inline index_t tindex( index_t i, index_t k = 0 ) const
+    {
+        GISMO_ASSERT(m_curElimId==0, "finalize() was not called on gsDofMapper");
+        if (m_tagged.empty())
+            return MAPPER_PATCH_DOF(i,k) + m_numCpldDofs - m_numFreeDofs;
+        else
+            return std::distance(m_tagged.begin(),std::lower_bound(m_tagged.begin(),m_tagged.end(),MAPPER_PATCH_DOF(i,k)));
```
filiatra: Shoud be just
return std::distance(m_tagged.begin(),std::lower_bound(m_tagged.begin(),m_tagged.end(),MAPPER_PATCH_DOF(i,k))); 

without if/else i think..

src/gsCore/gsDofMapper.h
```diff
@@ -346,9 +366,23 @@ class GISMO_EXPORT gsDofMapper
...
+    { return  is_tagged_index( index(i, k) ); }
+
+    /// Returns true if \a gl is a tagged dof.
+    inline bool is_tagged_index( index_t gl) const
+    {
+        if(m_tagged.empty())
+            return  (gl < m_numFreeDofs + m_shift                    ) && // is a free dof and
+                    (gl + m_numCpldDofs + 1 > m_numFreeDofs + m_shift);   // is not standard dof
+        else
+            return std::binary_search(m_tagged.begin(),m_tagged.end(),gl);
```
filiatra: Should be just "return std::binary_search(m_tagged.begin(),m_tagged.end(),gl);"

src/gsCore/gsDofMapper.h
DaHofa: Ups, you are right

src/gsCore/gsDofMapper.h
DaHofa: Ups, you are right

src/gsCore/gsDofMapper.cpp
DaHofa: Depends on what we store. m_tagged contains the index(i,k) of the tagged dofs.

DaHofa: Ahhh.. now i see what you mean. So i should store the index i, such that m_dofs[i] is tagged, isnt it? This would make the marking independent on the permutation.

src/gsCore/gsDofMapper.cpp
DaHofa: This is then related to the comment above

src/gsCore/gsDofMapper.cpp
DaHofa: wait, but if we only the the indices i, such that m_dofs[i] is tagged, then the calculation of is_tagged_index(..) and tindex(..) is more "complicated".  In order to search in for some gl in m_tagged, we first have to compute its preimage, i.e., the j, such that m_dofs[j] = gl and then do the search based on j in m_tagged...

filiatra: To be consistent, I think you should store the index(i,k) [global index].
ie.
for i=0.. m_dofs[size]
if ( is_coupled_index(m_dof[i])
m_tagged.push_back(m_dof[i]):

sort+remove duplicates before

(essentially you do these stuff in permuteFreeDofs, but it should be done already here..
note that:  m_numFreeDofs-m_numCpldDofs+i has nothing to do with index(i,k)...

filiatra: * Or actually, now that I see it again.. OK now I think your implementation is right here.

src/gsCore/gsDofMapper.cpp
filiatra: OK: this made me a little confused.
You could rename tagged --> tagged_permuted for clarity

src/gsCore/gsDofMapper.cpp
```diff
@@ -282,6 +295,42 @@ void gsDofMapper::setIdentity(index_t nPatches, size_t nDofs)
...
+    
+    for(index_t i=0; i<(index_t)dofs.size();++i)
+    {
+        index_t idx = dofs[i];
+        if(is_free_index(idx))
+        {
+            m_dofs[i] = permutation[idx];
+            //fill  bookkeeping for tagged dofs
+            if(is_tagged_index(idx))
+                tagged.push_back(m_dofs[i]);
```
filiatra: There is a small bug here for all tagged indices which are not free
(they should stay in "m_tagged" upermuted)

filiatra: Just add "else if is_tag_index(idx) tagged.push_back(idx);" to fix it

src/gsCore/gsDofMapper.cpp
DaHofa: ok

src/gsCore/gsDofMapper.cpp
DaHofa: ok

src/gsCore/gsDofMapper.cpp
DaHofa: uhh, thats fishy bug. nice.

DaHofa:
i will incorporate the changes on monday (maybe i have time on sunday, but i am not sure)

DaHofa:
so, this should work now, finally.

filiatra:
Thanx, you can now "Squash-and-merge"

DaHofa:
i will wait for the checks to be sure and then merge

PR #109, opened 2018-02-22
fixing gsplanardomain and gscurveloop

codewerfer:
* add another memory::make_unique implementation
* refactor gsPlanarDomain to return gsMatrix<T> instead raw pointer

### DIFF
--- a/src/gsCore/gsMemory.h
+++ b/src/gsCore/gsMemory.h
@@ -187,6 +187,12 @@ template <typename T>
 inline shared_ptr<T> make_shared_not_owned(const T *x)
 { return shared_ptr<T>(const_cast<T*>(x), null_deleter<T>); }
 
+
+#if __cplusplus >= 201402
+using std::make_unique;
+template <typename T>
+inline unique_ptr<T> make_unique(T * x) { return make_unique<T>(dynamic_cast<T*>(x)); }
+#else
 /// Takes a T* and wraps it in an unique_ptr. Useful for one-off
 /// function return values to avoid memory leaks.
 ///
@@ -198,6 +204,7 @@ inline unique_ptr<T> make_unique(T * x) { return unique_ptr<T>(x); }
 /// Takes a T*, cast it to C* and wraps it in an unique_ptr.
 template <class C, typename T>
 inline unique_ptr<C> make_unique(T * x) { return unique_ptr<C>(dynamic_cast<C*>(x)); }
+#endif
 
 /// \brief Converts a uPtr \a p to an uPtr
 /// of class \a C and gives it back as return value.
--- a/src/gsCore/gsMemory.h
+++ b/src/gsCore/gsMemory.h
@@ -187,12 +187,6 @@ template <typename T>
 inline shared_ptr<T> make_shared_not_owned(const T *x)
 { return shared_ptr<T>(const_cast<T*>(x), null_deleter<T>); }
 
-
-#if __cplusplus >= 201402
-using std::make_unique;
-template <typename T>
-inline unique_ptr<T> make_unique(T * x) { return make_unique<T>(dynamic_cast<T*>(x)); }
-#else
 /// Takes a T* and wraps it in an unique_ptr. Useful for one-off
 /// function return values to avoid memory leaks.
 ///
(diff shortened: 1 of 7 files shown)


src/gsCore/gsMemory.h
```diff
@@ -195,6 +195,10 @@ inline shared_ptr<T> make_shared_not_owned(const T *x)
 template <typename T>
 inline unique_ptr<T> make_unique(T * x) { return unique_ptr<T>(x); }
 
+/// Takes a T*, cast it to C* and wraps it in an unique_ptr.
+template <class C, typename T>
+inline unique_ptr<C> make_unique(T * x) { return unique_ptr<C>(dynamic_cast<C*>(x)); }
```
filiatra: what is the use of this ?
better commit it with some usage.
Also, dynamic_cast might fail and then we get a nullptr..

stefantakacs: i am also not happy with that; we should try not to hide *explicit* casts.

If c++ allows an implicit cast, everything works fine, so we can already write

make_unique<Base>( new Derived() );

codewerfer: Need a way to get a raw pointer from one type direct to anthers uPtr.

Happens in segment2d_domain_example.cpp#36:
gsBSpline<real_t>  b =*safe(dynamic_cast<gsBSpline<real_t> * > ( P->outer().singleCurve() ));

after refactor it would be something like:
gsBSpline<real_t>::uPtr b = memory::convert_ptr<gsBSpline<real_t> >( memory::make_unique(P->outer().singleCurve()) );

with this:
gsBSpline<real_t>::uPtr b = memory::make_unique<gsBSpline<real_t> >(P->outer().singleCurve());

stefantakacs: we do not yet have convert_ptr yet?

c++11 has some

std::static_pointer_cast
std::dynamic_pointer_cast
std::const_pointer_cast

i would follow them.

and if P->outer().singleCurve() would return a uPtr - which it should -, we could have

gsBSpline<real_t>::uPtr b = memory::dynamic_pointer_cast<gsBSpline<real_t> >(P->outer().singleCurve());

codewerfer: gsCurve<T> * gsCurveLoop<T>::singleCurve() const

would prefer a uPtr instead also.. should I do?

stefantakacs: i would say yes.

stefantakacs:
certainly, we should be able to decide if we want to static_cast or to dynamic_cast because they are highly non-equivalent.

stefantakacs:
moreover, i would have liked to split this up into 2 pull requests, one for gsMemory and one for other improvements, which could be merged faster

filiatra:
OK, so this PR can be renamed as fixing gsplanardomain and gscurveloop, without touching gsMemory

src/gsCore/gsMemory.h
codewerfer: > If c++ allows an implicit cast, everything works fine, so we can already write
> make_unique( new Derived() );
Isn't that nearly what std::make_unique (C++14) does?

stefantakacs: actually std::make_unique and std::make_shared are *completely* different to our memory::make_unique and memory::make_shared.

Ours take raw pointers and wrap them into the pointer type, std's functions also construct the object and take the constructor's arguments as their arguments.

see also http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

src/gsModeling/gsPlanarDomain.h
```diff
@@ -224,13 +224,13 @@ class gsPlanarDomain
 
     /// split this planar domain in two, returning the new planar domain created
     /// as a result.
-    gsPlanarDomain *split(int startIndex, int endIndex,
+    gsPlanarDomain::uPtr split(int startIndex, int endIndex,
```
filiatra: gsPlanarDomain::uPtr  --> uPtr

src/gsModeling/gsPlanarDomain.h
```diff
@@ -168,12 +168,11 @@ class gsPlanarDomain
         return true;
     };
 
-    /// \todo return non-pointer instead
-    gsMatrix<T> * boundingBox() const
```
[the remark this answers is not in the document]
codewerfer: done

PR #112, opened 2018-02-24
Small improvements

filiatra:
3 small improvements in 3 commits

### DIFF
--- a/src/gsIO/gsFileManager.cpp
+++ b/src/gsIO/gsFileManager.cpp
@@ -325,15 +325,15 @@ std::string gsFileManager::getFilename(std::string const & fn)
 void
 gsFileManager::open(const std::string & fn)
 {
-    int ret =
+
 #if defined(__APPLE__)
-        std::system( ("open " + fn + " &").c_str() );
+    const int ret = std::system( ("open " + fn + " &").c_str() );
 #elif defined(__linux__)
-        std::system( ("xdg-open " + fn + " &").c_str() );
+    const int ret = std::system( ("xdg-open " + fn + " &").c_str() );
 #elif defined(_WIN32)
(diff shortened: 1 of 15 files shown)


src/gsIO/gsFileManager.cpp
```diff
@@ -322,5 +322,24 @@ std::string gsFileManager::getFilename(std::string const & fn)
     return fn;
 }
 
+void
```
codewerfer: return a boolean - maybe implement it later, as we talked.

src/gsUtils/gsStopwatch.h
```diff
@@ -72,22 +72,28 @@ template <typename Clock>
...
     /// Return elapsed time in seconds.
-    double stop() const { return Clock::getTime() - m_start; }
+    double stop() { return m_value = Clock::getTime() - m_start; }
+
+    /// Returns the last recorded elapsed time
+    double elapsed() const { return m_value; };
 
     friend std::ostream& operator<< (std::ostream& os, const gsGenericStopwatch& sw)
-    { return formatTime(os, sw.stop()); }
+    { return formatTime(os, sw.m_value); }
```
codewerfer: I thought of something like:
gsStopWatch watch; // time: 0
... do something:
gsInfo << watch; // time: 0 + do something.
... do something other;
gsInfo << watch; // time : 0 + do something + do something other.
So << would get back the current elapsed time since start.

filiatra: I see, but we usually also need to store the timings..
What you propose will become

 gsStopWatch watch; // time: 0
... do something:
watch.stop();
gsInfo << watch; // time: 0 + do something.
... do something other;
watch.stop();
gsInfo << watch; // time : 0 + do something + do something other.

PR #113, opened 2018-02-26
Some improvements in mpi, and two new examples

rschneckenleitner:


src/gsMpi/gsMpiComm.h
```diff
@@ -141,6 +141,36 @@ class gsSerialComm
...
+     * @param[in] len The number of elements which will be received
+     * @param[in] source The rank of the process which sended the message
+     * @param[in] tag Specifies the message ID
+     */
+    template<typename T>
+    static int recv (T* out, int len, int source, int tag = 0)
+    {
+        return 0;
+    }
+
```
filiatra: I think Isend and Irecv should also be added here

rschneckenleitner: For Isend and Irecv I need additionally a MPI_Request structure as an argument.

filiatra: OK, for now just add
typedef int MPI_Request; 
in gsSerialComm.

This allows to add the functions in gsSerialComm as well

rschneckenleitner:
Ok, fine, thank you!
I will do the changes.

rschneckenleitner:
It works without this typedef. 
With this typedef I get an error.

If you are going to wrap MPI_Request in G+Smo then please also wrap MPI_Status.



Best Regards,
Rainer

src/gsMpi/gsMpiComm.h
```diff
@@ -141,6 +144,68 @@ class gsSerialComm
...
+     * @param[out] out The buffer to store the received data in
+     * @param[in] len The number of elements which will be received
+     * @param[in] source The rank of the process which sended the message
+     * @param[in] tag Specifies the message ID
+     */
+    template<typename T>
+    static int recv (T* out, int len, int source, MPI_Status status, int tag = 0)
+    {
+        return 0;
+    }
```
filiatra: This one is has a problem:
https://cdash.ricam.oeaw.ac.at/viewBuildError.php?onlydeltap&buildid=245103

PR #114, opened 2018-02-27
Robust single-patch preconditioners

stefantakacs:
This pr introduces:
* robust single-patch preconditioners with an interface that is based on ignoring the geometry function but which can be extended eg. to a rank 1 geometry approximation.
* the subspace corrected mass smoother
* unittests
* some additional stuff for the assemblers


src/gsCore/gsForwardDeclarations.h
```diff
@@ -63,199 +63,214 @@ enum gsNeedEnum
...
-template<class T = real_t> struct gsFieldCreator;
+template <class T=real_t>               struct gsFieldCreator;
 
 class gsOptionList;
 
-template<class T = real_t, int _Rows=-1, int _Cols=-1, 
-         int _Options  = 0|((_Rows==1 && _Cols!=1)?0x1:0)> class gsMatrix;
-template<class T = real_t, int _Rows=-1, int _Options = 0> class gsVector;
+template <class T=real_t, int _Rows=-1, int _Cols=-1,
+         int _Options  = 0|((_Rows==1 && _Cols!=1)?0x1:0)>
```
codewerfer: wow

stefantakacs: that's not new, i just made some white space changes...

stefantakacs:
It seems as if MSVS has problems linking together unittests if they use similar template instanciations:

https://cdash.ricam.oeaw.ac.at/viewBuildError.php?buildid=245541

I am not sure what to do.

filiatra:
You sould remome GISMO_EXPORT from templated classes..

stefantakacs:
thnks, this worked. this was a longstanding error, e.g., in gsBlockOp

src/gsMatrix/gsKronecker.hpp
```diff
@@ -0,0 +1,86 @@
...
+                    const index_t i = itA.row() * Br + itB.row(),
+                                  j = itA.col() * Bc + itB.col();
+                    result.insert(i,j) = itA.value() * itB.value();
+                }
+            }
+        }
+    }
+    
+    return result;
+}
```
filiatra: This function is copied from
https://eigen.tuxfamily.org/dox/unsupported/KroneckerTensorProduct_8h_source.html

It cannot be added to stable. We will add our own implementations of Kronecker product to gsTensor..

src/gsMatrix/gsKronecker.hpp
```diff
@@ -0,0 +1,86 @@
...
+    gsMatrix<T> result(Ar*Br, Ac*Bc);
+
+    for (index_t i=0; i < Ar; ++i)
+        for (index_t j=0; j < Ac; ++j)
+            for (index_t ii=0; ii < Br; ++ii)
+                for (index_t jj=0; jj < Bc; ++jj)
+                    result(i*Br+ii,j*Bc+jj) = A(i,j) * B(ii,jj);
+
+    return result;
+}
```
filiatra: Similarly this is restricted to 2 matrices..

src/gsSolver/gsKroneckerOp_.cpp
```diff
@@ -0,0 +1,21 @@
+/** @file gsKroneckerOp_.cpp
```
filiatra: no headers in *_.cpp files..

src/gsSolver/gsMatrixOp.h
```diff
@@ -126,9 +143,9 @@ class gsMatrixOp : public gsLinearOperator<typename MatrixType::Scalar>
   * \ingroup Solver
   */
 template <class Derived>
-typename gsMatrixOp<Derived>::uPtr makeMatrixOp(const Eigen::EigenBase<Derived>& mat, bool sym=false)
```
filiatra: "sym" flags make things complicated and I have received a lot of complaints for them.
Does it really make something a lot faster ? i.e., is it really needed ?

src/gsSolver/gsParameterDomainPreconditioners.h
```diff
@@ -0,0 +1,108 @@
...
+#include <gsSolver/gsLinearOperator.h>
+
+
+namespace gismo
+{
+
+/// @brief Provides preconditioners that live on the parameter domain.
+///
+/// This class provides efficient preconditioners based on the parameter domain,
+/// assuming to have \a gsTensorBasis.
```
filiatra: What does it mean that they "live on the parameter domain"? 

The assumption of tensorbasis should be enforced in the code

src/gsSolver/gsParameterDomainPreconditioners.hpp
```diff
@@ -0,0 +1,438 @@
...
+    gsMatrix<T> localMass;
+
+    // Make domain element iterator
+    typename gsBasis<T>::domainIter domIt = basis.makeDomainIterator();
+
+    // Set the number of integration points for each element
+    gsVector<index_t> numQuadNodes( basis.dim() );
+    for (int i = 0; i < basis.dim(); ++i)
+        numQuadNodes[i] = basis.degree(i) + 1;
+    domIt->computeQuadratureRule( numQuadNodes );
```
filiatra: domIt->computeQuadratureRule( numQuadNodes ); 
this function is deprecated and should not be used..

src/gsSolver/gsParameterDomainPreconditioners.h
```diff
@@ -0,0 +1,108 @@
...
+
+    // Helper functions for implementation, which might be of interest also for use in
+    // other functions
+    static gsSparseMatrix<T>                assembleMass            (const gsBasis<T>& basis);
+    static gsSparseMatrix<T>                assembleStiffness       (const gsBasis<T>& basis);
+    static std::vector< gsSparseMatrix<T> > assembleTensorMass      (const gsBasis<T>& basis);
+    static std::vector< gsSparseMatrix<T> > assembleTensorStiffness (const gsBasis<T>& basis);
+    static void handleDirichletConditions(gsSparseMatrix<T>& matrix, const gsBoundaryConditions<T>& bc,
+                                          const boxSide& west, const boxSide& east);
+
```
filiatra: I am not convinced of the necessity of all these functions, there is a lot of duplication with the assemblers..

src/gsSolver/gsParameterDomainPreconditioners.hpp
```diff
@@ -0,0 +1,438 @@
...
+    gsMatrix<T> localStiffness; // (dense) stiffness matrix within one grid cell
+
+    // Make domain element iterator
+    typename gsBasis<T>::domainIter domIt = basis.makeDomainIterator();
+
+    // Set the number of integration points for each element
+    gsVector<index_t> numQuadNodes( dim );
+    for (int i = 0; i < basis.dim(); ++i)
+        numQuadNodes[i] = basis.degree(i) + 1;
+    domIt->computeQuadratureRule( numQuadNodes );
```
filiatra: domIt->computeQuadratureRule( numQuadNodes ); 
is deprecated

src/gsAssembler/gsAssembler_.cpp
```diff
@@ -1,7 +1,16 @@
...
+    @brief Provides generic assembler routines
+
+    This file is part of the G+Smo library.
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Kleiss, A. Mantzaflaris, J. Sogn
+*/
```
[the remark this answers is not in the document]
stefantakacs: this seems to be quite nonuniform, but i can remove it from that file because its rather unrelated.

examples/poisson_example.cpp
```diff
@@ -56,87 +85,77 @@ int main(int argc, char *argv[])
...
-    bcInfo.addCondition(2, boundary::south, condition_type::neumann, &hSouth);
-    //! [Boundary conditions]
-
-    /*
-      //Alternatively: You can automatically create Dirichlet boundary
-      //conditions using one function (the exact solution) for all
-      //boundaries like this:
 
+    // Here, we just define Dirichlet boundary conditions everywhere,
+    // based on the made up solution
```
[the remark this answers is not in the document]
stefantakacs: because one can choose the domain and we have a made-up solution, so it is non-trivial to determine the "correct" boundary conditions.

in the following, commented out code and in the *.dox-file, the choice of neumann bc is explained.

src/gsAssembler/gsAssembler_.cpp
```diff
@@ -1,7 +1,16 @@
...
 
-#include <gsAssembler/gsAssembler.h>
```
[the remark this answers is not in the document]
stefantakacs: i do not see the point because the "*.h" file should always be included by the "*.hpp" file. so, the "_.cpp" has no need to do it again.

src/gsMatrix/gsKronecker.hpp
stefantakacs: it cannot be removed from this pr.

moving to gsTensor is possible. we can also add eigen's copyright information.

src/gsSolver/gsMatrixOp.h
stefantakacs: That's why I _remove_ those flags.

src/gsSolver/gsParameterDomainPreconditioners.h
stefantakacs: "live on the parameter domain" means that we ignore the contributions from the geometry transformation function.

src/gsMatrix/gsKronecker.hpp
stefantakacs: see above

src/gsSolver/gsParameterDomainPreconditioners.hpp
stefantakacs: i cannot find related information.

src/gsSolver/gsParameterDomainPreconditioners.hpp
stefantakacs: i cannot find related information.

src/gsSolver/gsParameterDomainPreconditioners.h
stefantakacs: they are light-weight approaches for the tensor-product case. here, without geometry, we can be very fast.

src/gsSolver/gsParameterDomainPreconditioners.h
stefantakacs: the tensor-bases do not have a common class, so it practically impossible to do so.

it will be even more impossible, if (what we are doing in devel) we want to apply such parameter-domain based preconditioners in a multi-patch manner because gsMultiBasis only gives us gsBasis. This means that the caller has the burden of casting around.

src/gsAssembler/gsAssembler_.cpp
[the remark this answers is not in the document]
filiatra: No, the .h should not be included in the .hpp

src/gsMatrix/gsKronecker.hpp
stefantakacs: it is possible to re-implement this from scratch, cf. 0379c44 
that, however, uses gsSparseEntries which needs 1 more copy and twice as much memory! so, it would be preferable if eigen's version - which is better - could be used.

stefantakacs:
it is possible to re-implement this from scratch, cf. 0379c44
that, however, uses gsSparseEntries which needs 1 more copy and twice as much memory! so, it would be preferable if eigen's version - which is better - could be used.

filiatra:
I would propose to split this PR in smaller ones.
In particular I am not convinced about adding the parameterdomain... and the kronecker functions in the current form..

stefantakacs:
Robust solvers based on the parameter domain #114 

That's the core of the pr

stefantakacs:
i do not see any problem with the kronecker products.

filiatra:
- We need kronecker products of more than 2 matrices
- computeQuadratureRule stuff should not be used and should be removed
- The parameterdomain assembler is somehting very adhoc/restrictive

These are the main issues I see..

stefantakacs:
* we do have kronecker products for any number of matrices
* computeQuadratureRule is nowhere declared as deprecated, nor there is any alternative mentioned
* parameterdomain assemblers are needed for the corresponding preconditioners, which usually work well and are heavily used in literature.

stefantakacs:
I have removed the assembling stuff for now; assembling is delegated to gsGenericAssembler.

stefantakacs:
The default constructor for gsMultiPatch does not work as expected. Is there a possibility to obtain a geometry map which is the identity?

src/gsSolver/gsParameterDomainPreconditioners.hpp
```diff
@@ -0,0 +1,325 @@
...
+        gsGenericAssembler<T> assembler(gsMultiPatch<T>(),gsMultiBasis<T>(basis),options,&bc);
+        result.push_back( assembler.assembleMass() );
+        return result;
+    }
+    else
+    {
+        result.reserve(d);
+        for ( index_t i=d-1; i!=-1; --i )
+        {
+            gsBoundaryConditions<T> local_bc = getBoundaryConditionsForDirection(bc,i);
```
stefantakacs: check ordering

stefantakacs:
Very nice.

There still a few places marked with todo. I will fix them prior merge. This will most likely only happen the week after next week.

PR #118, opened 2018-02-28
add one more unittest for gsMatrixOp [NOT MERGE]

stefantakacs:
This adds one more unittest for gsMatrixOp, which tends to fail sometimes.


unittests/gsMatrixOp_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+        A << 2,2,3,  4,5,6,  7,8,10;
+        gsMatrix<> B (3,3);
+        B << 1,4,7,  4,5,8,  7,8,10;
+
+        gsLinearOperator<>::Ptr Aop = makeMatrixOp(A.selfadjointView<Lower>());
+
+        A(0,0) = 1; // check that gsMatrixOp holds no copy
+
+        gsMatrix<> C;
+        Aop->toMatrix(C);
```
filiatra: On this line I get:

./bin/unittests  gsMat

                   G+Smo 
      Geometry plus Simulation modules
               version 0.8.3
Compiled by GNU GCC 6.4.0 (201103, glibc++ 20170704)
RICAM-Linz 2012 - 2017, http://gs.jku.at/gismo
* Start  gsMatrixOp_test:DenseMatrix.
  Finish gsMatrixOp_test:DenseMatrix in 0.00 seconds.
* Start  gsMatrixOp_test:DenseMatrixTransposed.
  Finish gsMatrixOp_test:DenseMatrixTransposed in 0.00 seconds.
* Start  gsMatrixOp_test:DenseMatrixSymm.
unittests: /home/samwise/amantzaflaris/workspace/trunk/stable/external/Eigen/src/Core/Product.h:97: Eigen::Product<Lhs, Rhs, Option>::Product(const Lhs&, const Rhs&) [with _Lhs = Eigen::SelfAdjointView<Eigen::Matrix<double, -1, -1>, 1u>; _Rhs = Eigen::Matrix<double, -1, -1>; int Option = 0; Eigen::Product<Lhs, Rhs, Option>::Lhs = Eigen::SelfAdjointView<Eigen::Matrix<double, -1, -1>, 1u>; Eigen::Product<Lhs, Rhs, Option>::Rhs = Eigen::Matrix<double, -1, -1>]: Assertion `lhs.cols() == rhs.rows() && "invalid matrix product" && "if you wanted a coeff-wise or a dot product use the respective explicit functions"' failed.
Aborted (core dumped)

stefantakacs: okey, i have to check that

stefantakacs: interestingly, this uses the _old_ constructor, i.e., that one i did not touch at all.

filiatra: I do not know what is going on, I also have to check more, because there is no error on cdash, can you reproduce it?

stefantakacs: yes; now its the same for me.

it seems that selfadjoinedView<Lower>() gives an Eigen expression that is somewhat different to the expression provided by transpose.

originally it did work for me and if i add
GISMO_ASSERT( rows()==cols(), "" );
to gsLinearOperator::toMatrix, it again works.

filiatra:
Two observations:
- The error with Eigen::SelfAdjointView is a bug in Eigen - we could report it
- The nw constructor by 2 arguments and the respective test It believe are buggy and will not run error free.

stefantakacs:
i somehow think its a problem with eigen.

stefantakacs:
close for now. i could not reproduce the problem without using gsMatrixOp.

PR #119, opened 2018-02-28
Remove sym from gsMatrixOp; add unittest

stefantakacs:
This now only consists of removing sym and unittests

### DIFF
--- a/src/gsSolver/gsMatrixOp.h
+++ b/src/gsSolver/gsMatrixOp.h
@@ -49,42 +49,40 @@ public:
     /// @note This does not copy the matrix. Make sure that the matrix
     /// is not deleted too early (alternatively use constructor by
     /// shared pointer)
-    gsMatrixOp(const MatrixType& mat, bool sym=false)
-    : m_mat(), m_expr(mat.derived()), m_symmetric(sym)
+    gsMatrixOp(const MatrixType& mat)
+    : m_mat(), m_expr(mat.derived())
     {
         //gsDebug<<typeid(m_expr).name()<<" Ref: "<<is_ref<NestedMatrix>::value<<"\n";
     }
 
     /// @brief Constructor taking a shared pointer
-    gsMatrixOp(const MatrixPtr& mat, bool sym=false)
-    : m_mat(mat), m_expr(m_mat->derived()), m_symmetric(sym)
+    gsMatrixOp(const MatrixPtr& mat)
+    : m_mat(mat), m_expr(m_mat->derived())
     { }
 
     /// @brief Make function returning a smart pointer
     ///
     /// @note This does not copy the matrix. Make sure that the matrix
     /// is not deleted too early or provide a shared pointer.
-    static uPtr make(const MatrixType& mat, bool sym=false)
-    { return memory::make_unique( new gsMatrixOp(mat,sym) ); }
+    static uPtr make(const MatrixType& mat)
+    { return memory::make_unique( new gsMatrixOp(mat) ); }
 
     /// Make function returning a smart pointer
-    static uPtr make(const MatrixPtr& mat, bool sym=false)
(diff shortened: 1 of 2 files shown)


src/gsSolver/gsMatrixOp.h
```diff
@@ -49,42 +49,40 @@ class gsMatrixOp : public gsLinearOperator<typename MatrixType::Scalar>
...
+    : m_mat(), m_expr(mat.derived())
     {
         //gsDebug<<typeid(m_expr).name()<<" Ref: "<<is_ref<NestedMatrix>::value<<"\n";
     }
 
     /// @brief Constructor taking a shared pointer
-    gsMatrixOp(const MatrixPtr& mat, bool sym=false)
-    : m_mat(mat), m_expr(m_mat->derived()), m_symmetric(sym)
+    gsMatrixOp(const MatrixPtr& mat)
+    : m_mat(mat), m_expr(m_mat->derived())
```
filiatra: I think that
gsMatrixOp(MatrixPtr mat) : m_mat(give(mat))

is more suitable here

src/gsSolver/gsMatrixOp.h
```diff
@@ -49,42 +49,40 @@ class gsMatrixOp : public gsLinearOperator<typename MatrixType::Scalar>
...
     /// is not deleted too early or provide a shared pointer.
-    static uPtr make(const MatrixType& mat, bool sym=false)
-    { return memory::make_unique( new gsMatrixOp(mat,sym) ); }
+    static uPtr make(const MatrixType& mat)
+    { return memory::make_unique( new gsMatrixOp(mat) ); }
 
     /// Make function returning a smart pointer
-    static uPtr make(const MatrixPtr& mat, bool sym=false)
-    { return memory::make_unique( new gsMatrixOp(mat,sym) ); }
+    static uPtr make(const MatrixPtr& mat)
```
filiatra: similarly, make(MatrixPtr mat)

src/gsSolver/gsMatrixOp.h
```diff
@@ -153,16 +150,16 @@ typename gsMatrixOp<Derived>::uPtr makeMatrixOp(const Eigen::EigenBase<Derived>&
...
+typename gsMatrixOp<Derived>::uPtr makeMatrixOp(const memory::shared_ptr<Derived> & mat)
 {
-    return memory::make_unique(new gsMatrixOp<Derived>(mat, sym));
+    return memory::make_unique(new gsMatrixOp<Derived>(mat));
 }
 
 // We need an additional guide for the compiler to be able to work well with unique ptrs
 template <class Derived>
-typename gsMatrixOp<Derived>::uPtr makeMatrixOp(memory::unique_ptr<Derived> mat, bool sym=false)
+typename gsMatrixOp<Derived>::uPtr makeMatrixOp(memory::unique_ptr<Derived> mat)
```
filiatra: This one should be possible to be removed, when the make(shared_ptr) is updated

stefantakacs:
We cannot merge this PR request as long as the issue with the unittest (cf. PR#118) is not fixed because this is needed in devel.

src/gsSolver/gsMatrixOp.h
stefantakacs: no, removing this does not work. here, the problem is not the move semantics but the missing ability of c++ to deduce "Derived".

src/gsSolver/gsMatrixOp.h
stefantakacs: done

src/gsSolver/gsMatrixOp.h
stefantakacs: done

stefantakacs:
found out that devel can now live without sym and without plugging selfadjoinedView into gsMatrixOp.

PR #120, opened 2018-02-28
poisson_example

stefantakacs:
This example shall also be an entry point for explaining our own preconditioners.

### DIFF
--- a/doc/poisson_example.dox
+++ b/doc/poisson_example.dox
@@ -4,29 +4,35 @@ namespace gismo {
 
 \page poisson_example poisson_example.cpp
 
-The Poisson equation: \f{eqnarray*}{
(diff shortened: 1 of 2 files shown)


filiatra:
This PR changes a little bit the style of this example. 
It was a plain verbose example setting up a particular domain, BCs etc..
It now becomes parameterized with many options and possibilities, ie. it changes radically its style.

There is the poisson2_example which has the intended style, ie. it is parameterized by several things and loads most options from a file (filedata/pde/poisson2d_bvp.xml). There eg. I could imagine to add an optionlist for the iterative solver. Overall, I think that including more options fits better there. What do you think?

stefantakacs:
that would be an option.

PR #123, opened 2018-03-01
better implementation of gsSparseMatrix::innerOf

filiatra:


filiatra:
Essentially to get a submatrix of a big sparse matrix without zero rows/or columns..

filiatra:
that is true, hopefully will be used in the future

PR #124, opened 2018-03-01
Update gsGridHierarchy.hpp

stefantakacs:
The current version gives 1 level more than requested. This is not a big deal but it would be good if it is consistent with buildByRefinement.


filiatra:
ok, please also have a look
https://cdash.ricam.oeaw.ac.at/viewBuildError.php?type=1&onlydeltap&buildid=246366

PR #127, opened 2018-03-07
gsMultiGridOp should take sparse matrices if available

stefantakacs:
The solver on the coarsest grid level should - if possible - be built on sparse matrix technology. This is at least for multipatch domains of interest, because there the coarsest grid problem might still be large-scale and sparse.

The previous implementation might stay there as fallback, but with a warning if the size is too large.


stefantakacs:
@filiatra There seems to be a problem with AppVeyor:

https://ci.appveyor.com/project/gismo/gismo/build/G+Smo-stable.477

filiatra:
yes, for some reason it tries to compile branches without PR..

PR #132, opened 2018-03-09
Bugfix in GMRES

DaHofa:
early exit in this function leads to a mismatch of matrix sizes later in gsGMRes<T>::finalizeIteration() with the effect that it crashes in line 49. 
Anyway I dont see a reason why this check is needed, the normal exit should be sufficient.


filiatra:
I agree, also the  hardcoded "1e-16" is not so nice, I let @stefantakacs approve it

DaHofa:
I have never seen such a check in other implementations, so I dont think its needed.

PR #133, opened 2018-03-09
gsKroneckerOp: Test + some explaination

stefantakacs:
It was good if we test that operator


filiatra:
Since the test is on integers it would be better to change the type to integer matrix and use CHECK_EQUAL(EXPECTED,ACTUAL);

PR #136, opened 2018-03-14
Enable partial and incremental installs

mmoelle1:
@filiatra: As discussed today, I am sending you a first version that implements individual install targets, i.e. install.gismo and install.gismo_dev. I have also added individual install targets for each example application so that you can easily test

make install.gsView
make install.CompressibleEuler (in the devel branch)

Once you are satisfied with the commit, you can remove them or we make an addition flag GISMO_WITH_INDIVIDUAL_INSTALL_TARGETS (or similar).

**[Update]** The current implementation also works under macOS if the variable DYLD_LIBRARY_PATH is set to <installdir>/lib so that the installed binary knows where to find the installed libraries.


filiatra:
Thanx Matthias (please call me "filiatra", since "angelos" is https://github.com/angelos :)

filiatra:
Hi,
Looking more into this problem I realize that the best way to have partial installations is
* Having set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY true) as it is already
* Making the executable targets optional (this way they are not installed if they were not built)

this way the dependencies are taken care of, and we do not have to do it manually (which can tricky)

The user should do:
1- make
2- make install

If the first make is not done the the install will fail. But is some examples were not built correctly it should still install only the examples that are compiled..

mmoelle1:
Did a fresh checkout with 'mkdir build; cd build; cmake ..' leaving all parameters to their default values.
`make gsView && make install`
gives the following error:
`...
[100%] Built target gsView
Install the project...
-- Install configuration: "Release"
CMake Error at cmake_install.cmake:36 (file):
  file INSTALL cannot find
  "/home/mmoller/codes/gismo.stable/build/lib/libgismo_static.a".
make: *** [install] Error 1
`
Tested under CentOS 7.

filiatra:
OK, this is now fixed (static libs should be optional), now partiall/incremental installs should work.

mmoelle1:
This works perfectly both with make and ninja build system. Please merge it into the stable/master branch.

PR #140, opened 2018-03-22
Parametrize the quadrature rule, add Gauss-Lobatto rules

filiatra:


src/gsAssembler/gsQuadRule.h
```diff
@@ -69,9 +57,21 @@ class gsQuadRule
      *
      */
     virtual void setNodes( gsVector<index_t> const & numNodes,
-                           unsigned digits = std::numeric_limits<T>::digits10 )
+                           unsigned digits = REAL_DIG )
```
stefantakacs: we might prfer c++ over c

src/gsAssembler/gsQuadrature.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /// Constructs a quadrature rule based on input \a options
+    template<class T>
+    static inline gsQuadRule<T> get(int qu, gsVector<index_t> const & numNodes)
+    {
+        switch (qu)
+        {
+        case GaussLobatto :
+            return gsLobattoRule<T>(numNodes);
+        default://GaussLegendre
```
stefantakacs: invalid values shoud invoke GISMO_ERROR

src/gsAssembler/gsQuadrature.h
```diff
@@ -0,0 +1,82 @@
...
+        const int qu = options.askInt("quRule", GaussLegendre);
+        const T       quA = options.getReal("quA");
+        const index_t quB = options.getInt ("quB");
+        const gsVector<index_t> nnodes = numNodes(basis,quA,quB,fixDir);
+        return get<T>(qu, nnodes);
+    }
+
+    /// Constructs a quadrature rule based on input \a options
+    template<class T>
+    static inline gsQuadRule<T> get(int qu, gsVector<index_t> const & numNodes)
```
stefantakacs: why "int"? shouln't it be "rule"?

src/gsAssembler/gsQuadrature.h
```diff
@@ -0,0 +1,82 @@
...
+            return gsGaussRule<T>(numNodes);
+        };
+    }
+
+    /// Computes and integer quA*deg_i + quB where deg_i is the degree
+    /// of \a basis
+    template<class T>
+    static gsVector<index_t> numNodes(const gsBasis<T> & basis,
+                               const T quA, const int quB, int fixDir = -1)
+    {
```
stefantakacs: why not index_t?

many followups...

src/gsAssembler/gsQuadRule.h
filiatra: This is defined in gsMath.h and is equal to std::numeric_limits<T>::digits10

src/gsAssembler/gsQuadrature.h
filiatra: GISMO_ERROR will be invoked for invalid values, so passing int 1 or 2 is perfectly fine, and I want ot allow it, as it is the same as the enum

src/gsAssembler/gsQuadrature.h
filiatra: I want to make sure that -1 is a valid value (currently index_t=int but we can imagine index_t=unsigned)

src/gsAssembler/gsQuadrature.h
stefantakacs: This seems to be a general question. Doing the same, e.g., for InterfaceStrategy and BoundaryStrategy in assemblers and dof mapper etc. would also save many casts...

filiatra: yes, in fact in the assemblers these will go away completely, they are in the options instead..

PR #141, opened 2018-03-23
add patchSide::getContainedCorners (int dim, std::vector<patchCorner> &corners)

stefantakacs:
We can ask a patchCorner for the coressponding patchSides, but we cannot ask a patchSide for the corresponding patchCorners.

This commit fixes that.

### DIFF
--- a/src/gsCore/gsBoundary.cpp
+++ b/src/gsCore/gsBoundary.cpp
@@ -32,6 +32,16 @@ void boxSide::getContainedCorners (int dim, std::vector<boxCorner> &corners) con
     }
 }
 
+void patchSide::getContainedCorners (int dim, std::vector<patchCorner> &corners) const
+{
+    std::vector<boxCorner> tmp;
+    boxSide::getContainedCorners(dim, tmp);
+    corners.resize(0);
+    corners.reserve(tmp.size());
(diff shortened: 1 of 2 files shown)


src/gsCore/gsBoundary.cpp
```diff
@@ -32,6 +32,16 @@ void boxSide::getContainedCorners (int dim, std::vector<boxCorner> &corners) con
     }
 }
 
+void patchSide::getContainedCorners (int dim, std::vector<patchCorner> &corners) const
+{
+    std::vector<boxCorner> tmp;
+    boxSide::getContainedCorners(dim, tmp);
+    corners.resize(0);
```
filiatra: corners.resize(0);  --->   corners.clear();

src/gsCore/gsBoundary.h
```diff
@@ -240,6 +242,15 @@ struct patchSide: public boxSide
     boxSide& side()       {return *this;}
     const boxSide& side() const {return *this;}
 
+    /**
+     * @brief returns the vector of the corners contained in the side
+     * @param dim is the ambient dimension
+     * @param corners
+     */
+    void getContainedCorners (int dim, std::vector<patchCorner> &corners) const;
```
filiatra: This function (and the existing overload) should be static

PR #142, opened 2018-03-26
add patchSide::getContainedCorners (int dim, std::vector<patchCorner> &corners)

stefantakacs:
fix linkage


stefantakacs:
I somewhat managed not to push this commit prior merge...

PR #144, opened 2018-03-26
Update ctest scripts for more automation

filiatra:


cmake/ctest_script.cmake
```diff
@@ -55,53 +72,55 @@ set(CTEST_CMAKE_GENERATOR "Unix Makefiles")
...
+exec_program(${HOSTNAME_CMD} ARGS OUTPUT_VARIABLE HOSTNAME)
+set(CTEST_SITE "${HOSTNAME}")
+endif(NOT HOSTNAME)
 
-# Source folder
-set(CTEST_SOURCE_DIRECTORY ${CTEST_SCRIPT_DIRECTORY}/..)
+# Name of this build
+if(NOT CTEST_BUILD_NAME)
+find_program(UNAME NAMES uname)
+exec_program("${UNAME}" ARGS "-s" OUTPUT_VARIABLE osname)
```
codewerfer: deprecated, execute_process should be used instead:
https://cmake.org/cmake/help/v3.8/command/exec_program.html

PR #145, opened 2018-03-26
access to coordinates of patchCorners and to midpoints of patchSides

stefantakacs:


stefantakacs:
It might be possible to provide the same for whole patches:

+    /// Get coordinates of the patch \a p in the parameter domain
+    void coordinates( index_t p, gsMatrix<T>& coordinates );
+
+    /// Get coordinates of the patch \a p in the physical domain
+    void physicalCoordinates( index_t p, gsMatrix<T>& coordinates );

filiatra:
This exists in gsGeometry::parameterCenter()

src/gsCore/gsMultiPatch.hpp
```diff
@@ -240,6 +240,35 @@ void gsMultiPatch<T>::addInterface( gsGeometry<T>* g1, boxSide s1,
...
+
+template<class T>
+void gsMultiPatch<T>::coordinates( const patchSide& ps, gsMatrix<T>& coordinates )
+{
+    coordinates.resize(m_dim,1);
+    gsMatrix<T> supp = m_patches[ps.patch]->parameterRange();
+    const index_t dir = ps.direction();
+    for (index_t d=0; d<m_dim;++d)
+        coordinates(d,0) = d == dir ? ( ps.parameter() ? supp(d,1) : supp(d,0) )
+                                    : ( ( supp(d,1) + supp(d,0) ) / 2 );
```
filiatra: please avoid the ternary operator in this specific case and use "if" instead (arithmetic operations can cause compilation problems with non-standard number types)

stefantakacs:
I think gsGeometry could give centers for boxCorners and boxSides but for patchCorners and patchSides we need to do it in gsMultiPatch.

stefantakacs:
>/// Return a point the side \a ps (corresponding to a central parametric value)
>gsMatrix pointOn(const patchSide& ps)

This is not very appropriate for corners.

src/gsCore/gsMultiPatch.h
```diff
@@ -226,12 +226,18 @@ class gsMultiPatch : public gsBoxTopology, public gsFunctionSet<T>
...
+    /// Add side \a s of patch \a g to the outer boundary of the domain
     void addPatchBoundary( gsGeometry<T>* g, boxSide s ) {
         int p = findPatchIndex( g );
         gsBoxTopology::addBoundary( patchSide( p, s ) );
     }
 
+    /// Get coordinates of the patchCorner \a pc in the physical domain
+    gsMatrix<T> pointOn( const patchCorner& pc );
+
+    /// Get coordinates of the midpoint of the the patchSide \a ps in the physical domain
```
filiatra: "of the midpoint of the"  --> is misleading, maybe "of a 'central' point on the patchSide"
the the --> the

PR #148, opened 2018-04-04
Remove add patch

codewerfer:
removed gsMultiPatch::addPatch( gsGeometry<T>* )

### DIFF
--- a/src/gsHSplines/gsTHBSplineBasis.hpp
+++ b/src/gsHSplines/gsTHBSplineBasis.hpp
@@ -528,8 +528,7 @@ gsMultiPatch<T> gsTHBSplineBasis<d,T>::getBsplinePatchesToMultiPatch(const gsMat
         p2 = b2.row(i).transpose();
 
         this->getBsplinePatchGlobal(p1, p2, level[i], geom_coef, temp1, cku, ckv);
-        typename gsTensorBSpline<2, T>::uPtr tbspline(new gsTensorBSpline<2, T>(cku, ckv, give(temp1)));
-        result.addPatch(tbspline);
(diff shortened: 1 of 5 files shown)


src/gsHSplines/gsTHBSplineBasis.hpp
```diff
@@ -528,8 +528,7 @@ gsMultiPatch<T> gsTHBSplineBasis<d,T>::getBsplinePatchesToMultiPatch(const gsMat
         p2 = b2.row(i).transpose();
 
         this->getBsplinePatchGlobal(p1, p2, level[i], geom_coef, temp1, cku, ckv);
-        typename gsTensorBSpline<2, T>::uPtr tbspline(new gsTensorBSpline<2, T>(cku, ckv, give(temp1)));
-        result.addPatch(tbspline);
+        result.addPatch(typename gsTensorBSpline<2, T>::uPtr(new gsTensorBSpline<2, T>(cku, ckv, give(temp1))));
```
stefantakacs: result.addPatch(gsTensorBSpline<2, T>::make(cku, ckv, give(temp1)));

does not work!?

codewerfer: Not sure where make should came from. gsTensorBSpline<d,T> doesn't have that member, nor it's base classes (or do I oversee it?)

codewerfer: Looking into gsMultiGrid.h and gsPreconditioner.h, should we put a make function to gsFunctionSet.h with same working like the clone function? (using the Macro such that every derived class has this make, with it's own implementation)

PR #150, opened 2018-04-10
Update version handling

filiatra:


cmake/gsLibrary.cmake
```diff
@@ -39,7 +39,7 @@ endif()
   set_target_properties(${PROJECT_NAME} PROPERTIES
```
codewerfer: CMake Error at stable/cmake/gsLibrary.cmake:39 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
Call Stack (most recent call first):                                                                                                                                                                                                         
  devel/CMakeLists.txt:139 (include)

PR #177, opened 2018-05-18
fix minor issues

filiatra:


filiatra:
It is bad practice to call exit(), and we need to elimintate it. Instead, for more beautiful message one can handle the exception as in the commit for gsView..

external/tclap/CmdLine.h
```diff
@@ -504,7 +504,8 @@ inline void CmdLine::parse(std::vector<std::string>& args)
 	}
 
 	if (shouldExit)
-		exit(estat);
+        throw(CmdLineParseException("Command line argument exception", gismo::util::to_string(estat))); // G+Smo
```
stefantakacs: It is supposed to be a clean exit; we might consider (re-)throwing ExitException.

stefantakacs:
// add to gsCmdLine.h

class gsExitException {
    int exit_code;
public:
    gsExitException(int _exit_code) : exit_code(_exit_code) {}
    operator int ()    { return exit_code;         }
    const char* what() { return "gsExitException"; }
};

// callers invoke:

try { cmd.getValues(argc,argv); } catch( gsExitException e ) { return e; }

Then we could even preserve the current behavior that --help and --version yield
exit code 0.

stefantakacs:
@filiatra it is up to you if you want to merge this or a previous version

stefantakacs:
any thoughts about this pr?

filiatra:
I changed it to just through the exit status, look fine to me now

stefantakacs:
Yes, this is also a good solution.

I will update devel after you have merged this.

PR #179, opened 2018-05-23
smaller fixes due to clang-tidy

stefantakacs:


src/gsCore/gsConstantFunction.h
```diff
@@ -58,37 +58,37 @@ class gsConstantFunction : public gsGeometry<T>
         : m_domainDim(domainDim)
     {
         m_coefs.resize(1,1);
-        m_coefs.at(0) = x;
```
filiatra: why is this better ?

stefantakacs: its a non-standard extension to the eigen interface and i did not find in the internet what it exactly does until i found it it gismo

we can keep the original if you prefer. the main thing was to fix line 91:

m_coefs.at(2) = w;

which should be 3.

filiatra: ok, I see. The at() is a little faster, because it does not need to compute the position, eg j*n+i, but this is really not important!

PR #181, opened 2018-05-25
fix error in gsBasis

stefantakacs:
clang claims:

src/gsCore/gsBasis.hpp:438:1: warning: all paths through this function will call itself [-Winfinite-recursion]

and I concur.

I do not know if this is the best possible fix.


stefantakacs:
It seems as if most of the derived classes overwrite this function. So, we might also consider GISMO_NO_IMPLEMENTATION

Alternatively, we can force this function to be used by marking it final and remove the overwrites in the derived classes...

filiatra:
The fix is OK, it is complicated to further change the derived classes due to return value..

stefantakacs:
Yes, indeed. That's why I have fixed it that way...

PR #184, opened 2018-05-29
Enable support of multiple exprtk adaptors (helpful if G+Smo is used in header-only mode)

mmoelle1:
This commit makes it possible to use multiple export adaptors at the same time.
This is particularly helpful of G+Smo is used as header-only library, when one might want to use the standard adaptor and mix them with adaptors for GMP/MPFR.

This commit is a preparation for committing support for Universal NUMbers (UNUMs).


src/gsCore/gsFunctionExpr.hpp
```diff
@@ -52,23 +52,31 @@
...
+#include <exprtk_mpfr_adaptor.hpp> // external file
+#endif
+
+#if defined(GISMO_WITH_MPQ)
+#include <exprtk_gmp_adaptor.hpp>  // external file
 #endif
 
+#if defined(GISMO_WITH_CODIPACK)
+#include <gsCoDiPack/exprtk_codi_rf_adaptor.hpp>
+#include <gsCoDiPack/exprtk_codi_rr_adaptor.hpp>
```
filiatra: double iinclusion!

filiatra: ok, I did not see the rr/rf :)

filiatra:
Seems that constants are always the same (doubles) , how about moving it to a file exprtk_constants.h  ?

mmoelle1:
No, they are often of type double but, GMP/MPFR (and later Unum) use different types.

filiatra:
As far as I can see GMP used double, maybe also for Unum is enough ? I dont know!

PR #191, opened 2018-06-05
Fixed bug when mixing GISMO_WITH_MPQ/GISMO_WITH_MPFR with other data types

mmoelle1:
G+Smo did not support enabling GISMO_WITH_MPQ/GISMO_WITH_MPFR together
with order data types, such as, GISMO_WITH_UNUM. Moreover, it was not
possible to enable GISMO_WITH_MPQ and GISMO_WITH_MPFR without setting
GISMO_COEFF_TYPE to mpq_class and mpfr::real, respectively.

This is now possible with this commit. If G+Smo is compiled in
header-only mode, one can now perform all computations using, say,
doubles but enable MPQ/MPFR support to validate the result.

The main change introduced with this commit it the separation of
exprtk_X_adaptor.hpp into the two files

exprtk_X_forward.hpp

and

exprtk_X_adaptor.hpp

The forward declaration file only contains specializations of
is_true(), is_false(), etcetera. The adaptor file provides the actual
implementation. This splitting was implicitly done in the old version.

Note: This commit does not change legacy behavior, that is, only
setting GISMO_COEFF_TYPE to mpq_class or mpfr::real and letting G+Smo
enable the ncessary GISMO_WITH_MPQ/GISMO_WITH_MPFR flags internally.


filiatra:
O kI wll, in posit, there is one constructor missing:

    posit(const long initial_value)          { *this = initial_value; }

this causes the ambiguous overload...

PR #196, opened 2018-06-13
iterative fitting method

johannaMM:
the iterative fitting method I presented at the g+smo developer days, implemented into the class gsTensorFitting, including an example tensor_grid_fitting_example.cpp and data in filedata/fitting/grid_surface.xml


src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+	 * including a fitting step, a knot removing step and a projection step.
+	 **/ 	
+	
+	template<class T>
+	class gsTensorFitting
+	{
+		public:
+		
+		/// default constructor
+		gsTensorFitting(){}
```
osmihran, line 33: Do we need it?

src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+		/// returns the meanPointError; throws if .compute() has not been called yet
+		T meanPointError() const {return mean_error_; }
+		
+		/// returns the maxPointError; throws if .compute() has not been called yet
+		T maxPointError() const {return max_error_; }
+		
+		///computes the BSplineSurface
+		void compute(); 
+		
+		private:
```
osmihran, line 84: Is there a specific reason to use `private` and not `protected`?

src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+		* @param[out] u_k	size = n_p and filled with 0s required, containing n_p parameter values
+		*/
+		void param_vals_centripetal(const int dir, std::vector<T>& u_k);
+		
+		/** constructs knot vector out of the values contained in another vector 
+		* @param[in] vec	contains knot-values u_0, .., u_m
+		* @param[in] m	size of vec
+		* @return knot vector with 0, 0, u_1, .., u_m-1, 1, 1
+		*/ 
+		gsKnotVector<T> get_knotvec_from_vec(const std::vector<T>& vec, const int m);
```
osmihran, line 110: Don't we already have a very similar function in `gsKnotVector`?

src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+		/** searches maximum of all entries in matrix M
+		* @param[in] M	matrix
+		* @param[out] _row	row-index of maximum
+		* @param[out] _col	column-index of maximum
+		* @param[out] _value	maximum
+		*/ 
+		void get_max_in_matrix(const gsMatrix<T>& M, index_t& _row, index_t& _col, T& _value); 
+		
+		//----------------------------------------------------------------------
+		//members
```
osmihran, line 322: We should stick to the G+Smo convention and call them `m_<name>`.

src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+		
+		/** projects a given point Q onto the surface defined by Tensor, using (u0,v0) as a starting point
+		* @param[in] u0	starting point in one direction
+		* @param[in] v0	starting point in other direction
+		* @param[in] Tensor	B-Spline surface, needed for evaluation of points and first and second derivatives
+		* @param[in] Q	point (as one row of a matrix) which gets projected onto a surface
+		* @param[out] uk	parameter value in one direction which yields Q
+		* @param[out] vk	parameter value in other direction which yields Q
+		*/
+		void project_to_surface(const T& u0, const T& v0, const gsTensorBSpline<2,T>& Tensor, const gsMatrix<T>& Q, T& uk, T& vk);
```
osmihran, line 296: This might be worth generalising from 2 to `d` and moving to `gsTensorBSpline`.

src/gsTensor/gsTensorFitting.h
```diff
@@ -0,0 +1,355 @@
...
+		* @param[in] eps2	epsilon for zero cosine check
+		* @return bool-variable which is true if either point coincidence or zero cosine is satisfied
+		*/
+		bool check_conv_crit(const gsMatrix<T>& R, const std::vector<T>& k, const std::vector<T>& Su, const std::vector<T>& Sv, const T& eps1, const T& eps2); 
+		
+		/** checks if point uk is in range [0,1]
+		* if not, then uk is set to 0 resp. 1
+		* @param[in,out] uk
+		*/
+		void check_new_vals(T& uk);
```
osmihran, line 276: Maybe rename? It is not really checking (which would probably return a `bool`) but setting.

osmihran:
Recently I've been thinking about this PR quite often. When the new submodule system is productive, we can consider moving this into a submodule; currently I don't think it should go into stable.

PR #223, opened 2018-08-28
Gs geometry evaluator2

codewerfer:
moved gsGeometryEvaluator to devel
moved gsNorms to devel
improved gsAssembler
improved gsVisitor*


src/gsAssembler/gsAssembler.h
```diff
@@ -34,6 +34,236 @@
...
+            break;
+    }
+}
+
+template<typename T>
+void secDerToTensor(Eigen::DenseBase<Eigen::Map<const Eigen::Matrix<double, -1, -1>, 0, Eigen::Stride<0, 0> > >::ConstColXpr & secDers,
+                    gsMatrix<T> * a,
+                    int parDim, int geoDim)
+{
+    static const int dim = parDim * (parDim + 1) / 2;
```
stefantakacs: why static?

src/gsAssembler/gsAssembler.h
```diff
@@ -34,6 +34,236 @@
...
+void transformDeriv2Hgrad(const gsMapData<T> & md, 
+                          index_t              k,
+                          const gsMatrix<T> &  funcGrad,
+                          const gsMatrix<T> &  funcSecDir,
+                          gsMatrix<T> &        result)
+{
+    //todo: check me
+    static const int ParDim = md.dim.first;
+    static const int GeoDim = md.dim.second;
+    GISMO_ASSERT(
```
stefantakacs: why static?

src/gsAssembler/gsAssembler.h
```diff
@@ -34,6 +34,236 @@
...
+        (ParDim == 1 && (GeoDim == 1 || GeoDim == 2 || GeoDim == 3))
+        || (ParDim == 2 && (GeoDim == 2 || GeoDim == 3))
+        || (ParDim == 3 && GeoDim == 3), "No implementation for this case");
+
+    // important sizes
+    static const int parSecDirSize = ParDim * (ParDim + 1) / 2;
+    static const int fisSecDirSize = GeoDim * (GeoDim + 1) / 2;
```
stefantakacs: why static?

src/gsAssembler/gsAssembler.h
```diff
@@ -34,6 +34,236 @@
...
+    {
+        secDerToHessian<T>(funcSecDir.block(i * parSecDirSize, k, parSecDirSize, 1), parFuncHessian, ParDim);
+        hessianToSecDer<T>(JM1 * parFuncHessian * JMT, result.row(i), GeoDim);
+    }
+
+    // Second part: sum_i[  J^-T H(G_i) J^-1 ( e_i^T J^-T grad(u) ) ]
+    const gsAsConstMatrix<T, -1, -1>  & secDer = md.deriv2(k);
+#ifndef _MSC_VER
+    gsMatrix<T> DDG[GeoDim]; // Each matrix is the Hessian of a component of the Geometry
+#else
```
stefantakacs: can you use the STACK_ARRAY macro here?

certainly, this is not allowed in standard c++ because GeoDim cannot be known at compile time.

in every case, it would be good to encapsulate such constructs and give the implementation in gsCore/gsMemory.h

If this is done on the heap, consider to use memory::unique_ptr

codewerfer: nope - doesn't work.

./biharmonic_example 
Assembling...
double free or corruption (out)
Aborted (core dumped)

For the moment we should keep it as it is..

codewerfer: maybe we should move all of this functions to gsFuncData or gsMapData.. than we could use template parameters again - inverse() is way faster in such a case (knows ParDim at compile time), or?

PR #228, opened 2018-09-21
update to eigen 3.3.5

filiatra:


filiatra:
will merge a newer version of eigen

PR #233, opened 2018-10-04
fixed a bug in normal()

DaHofa:
dimensions were not agreeing when calling md.jacobians().block(..). Led to a crash in gsShellMixed_test (devel).


filiatra:
thanx!

PR #243, opened 2018-11-08
Allow to build only one of the subprojects

stefantakacs:
I personally prefer not to be forced to build motor.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,11 +370,12 @@ if (NOT gismo_optionals STREQUAL "")
   message(STATUS "Optional compile list: ${gismo_optionals}")
 endif()
 
-if(GISMO_UNSUPPORTED)
+if(GISMO_UNSUPPORTED OR GISMO_MOTOR)
   gismo_fetch_module(unsupported)
-  if(NOT GISMO_NO_MOTOR)
-    gismo_fetch_module(motor)
-  endif()
+endif()
+
+if(GISMO_MOTOR)
+  gismo_fetch_module(motor)
 endif()
 
 unset(GISMO_UNAME CACHE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,7 +370,7 @@ if (NOT gismo_optionals STREQUAL "")
   message(STATUS "Optional compile list: ${gismo_optionals}")
 endif()
 
-if(GISMO_UNSUPPORTED)
+if(GISMO_UNSUPPORTED OR GISMO_MOTOR)
   gismo_fetch_module(unsupported)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -370,12 +370,11 @@ if (NOT gismo_optionals STREQUAL "")
   message(STATUS "Optional compile list: ${gismo_optionals}")
 endif()
 
-if(GISMO_UNSUPPORTED OR GISMO_MOTOR)
+if(GISMO_UNSUPPORTED)
   gismo_fetch_module(unsupported)
-endif()
-
-if(GISMO_MOTOR)
-  gismo_fetch_module(motor)
+  if(NOT GISMO_NO_MOTOR)
+    gismo_fetch_module(motor)
+  endif()
 endif()
 
 unset(GISMO_UNAME CACHE)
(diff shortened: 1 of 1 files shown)


filiatra:
let's keep this in mind, and merge when cronjob nightly scripts are updated..
Also, motor most likely depends on unsupported..

stefantakacs:
Yes, we are not in a hurry.

stefantakacs:
@codewerfer Please add GISMO_MOTOR=ON where GISMO_UNSUPPORTED=ON within server scripts.

mmoelle1:
Let me briefly react on this. The MOTOR project is over and the number of MOTOR partners that continue developing in G+Smo is moderate. It’s basically UWB Pilzen, JKU, TU Kaiserslautern and TU Delft. The two latter ones already know about the new layout of G+Smo extensions. JKU, that’s you and UWB just needs one mail to three people (Bohumir, Hana, and Eva) and they can update their checkouts.

codewerfer:
OK, so revert my last commit.

stefantakacs:
@codewerfer  Did you update the scripts?

stefantakacs:
Btw, is there a possibility to send out an email via github?

stefantakacs:
I now suggest some rearrangement which make the file clearer.

I hope this still works!?

codewerfer:
sorry, had to do some support.
No, not finished with all servers yet - only my machine.

stefantakacs:
Please write me here, when you are done

codewerfer:
Sorry, https://github.com/gismo/gsElasticity/issues/10 costs my some time.
It is similar to this one - and that what we talked about how to prevent to clone **motor** submodule.

Btw. you can squash and merge this branch. It will not build motor on machines where it is not set yet, but the build will fail on most of this machines because of https://github.com/gismo/gsElasticity/issues/10 anyway.

CMakeLists.txt
```diff
@@ -306,14 +306,6 @@ if(GISMO_BUILD_LIB)
...
 endif()
 
-if(GISMO_ELASTICITY)
-  gismo_fetch_module(gsElasticity)
-endif()
-
-if(GISMO_EXASTENCILS)
-  gismo_fetch_module(gsExaStencils)
-endif()
-
```
filiatra: I think there was a reason why this was in this position, does everything still work after this commit ?

stefantakacs: I have tried to invoke
```
cmake . -DGISMO_ELASTICITY=On
cmake . -DGISMO_EXASTENCILS=On
make
```
in my build folder and it worked (up to the bugs the code had/has anyway).

CMakeLists.txt
```diff
@@ -381,10 +373,25 @@ if (NOT gismo_optionals STREQUAL "")
...
 endif()
 
+if(GISMO_ELASTICITY)
+  gismo_fetch_module(gsElasticity)
+endif()
+
+if(GISMO_EXASTENCILS)
+  gismo_fetch_module(gsExaStencils)
+endif()
+
```
codewerfer: @filiatra this code you mentioned is still there

stefantakacs: its about the order. that might matter.

stefantakacs:
Can it be true that (after downloading the submodule gsElasticity) that submodule is included even if it is deactivated in cmake?

Maybe because its name starts with gs and there is a corresponding rule?

codewerfer:
> Can it be true that (after downloading the submodule gsElasticity) that submodule is included even if it is deactivated in cmake?

Yes. That's the problem we get on @rschneckenleitner's computer.
Workaround to solve this:
`git submodule deinit extensions/gsElasticity`
 
> Maybe because its name starts with gs and there is a corresponding rule?

I do not know. If so, it will be also with `gsExaStencils`.
https://github.com/gismo/gismo/blob/5e83df1d3180291700b66e80a2209a65f593b6b0/cmake/gismoUse.cmake#L78-L88
A better solution for this would be nice.

----
If changed the wiki therefore, that somebody should `git clone <gismo>` without `--recusrive-submodules` and activate - and clone - them via the `-DGISMO_<submodule>` option.

codewerfer:
> Please write me here, when you are done

@stefantakacs Have updated all cdash servers I have access to.
Have `submodule deinit` all submodules. Have added `-DGISMO_MOTOR=ON` to build script.

stefantakacs:
Then let's try this.

PR #258, opened 2018-12-05
Preconder updates

stefantakacs:
This adds some more unittests (such that all gsPatchPreconditionerCreator functions are covered) and fixes some of those functions (concerning the ordering of the dimensions in the Kronecker products).

### DIFF
--- a/src/gsSolver/gsPatchPreconditionersCreator.hpp
+++ b/src/gsSolver/gsPatchPreconditionersCreator.hpp
@@ -106,8 +106,8 @@ std::vector< gsSparseMatrix<T> > assembleTensorMass_impl(
     std::vector< gsSparseMatrix<T> > result(d);
     for ( index_t i=0; i!=d; ++i )
     {
-        result[i] = assembleMass(tb->component(d-i-1));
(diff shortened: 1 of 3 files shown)


src/gsSolver/gsPatchPreconditionersCreator.hpp
```diff
@@ -324,10 +324,10 @@ typename gsPatchPreconditionersCreator<T>::OpUPtr gsPatchPreconditionersCreator<
     gsMatrix<T> ev;
 
     // Now, setup the Q's and update the D's
-    for ( index_t i=0; i<d; ++i )
+    for ( index_t i=d-1; i>=0; --i )
     {
         // Solve generalized eigenvalue problem
-        ges.compute(local_stiff[i], local_mass [i], Eigen::ComputeEigenvectors);
+        ges.compute(local_stiff[i], local_mass[i], Eigen::ComputeEigenvectors);
```
codewerfer: `d-1-i` or `i`?

stefantakacs: yes, indeed. i fix this a bit differently.

PR #260, opened 2018-12-10
[NOT MERGE] Allow to get transfer matrices for patches and for globs

stefantakacs:
This allows to decompose the patches into globs and to obtain local bases.

This is needed for the robust multi-patch multigrid stuff.


stefantakacs:
@filiatra This allows to decompose the dofs into globs.

There might be alternative names.

So far, this does not work as I want for the dG case. So, some rewrite is necessary.

But it would be good to agree on the interface.

stefantakacs:
I think I have a better idea. Close for now

PR #276, opened 2018-12-17
Allow to decompose a box into its components; add one example in multigrid

stefantakacs:
This pr contains a new generalization of boxSide and boxCorner: boxComponent.

This can be the interior, one face, one edge or one corner, etc.

Analogously to patchSide and patchCorner, we have patchComponent.

Now, one can query a gsBoxTopology for all its components (where "the same" components are combined into one).

Then, one can query a basis or a multibasis for the basis for the particular component and/or for the associated indices.

The changes to multiGrid_example show that the new code is useful.


stefantakacs:
I would like to cite the new multigrid examples in one paper that I would submit within the next few days. (Hopefully.)

stefantakacs:
Unittests are still missing.

I would add them if we would agree on the interfaces.

stefantakacs:
I hope the diff in gsMultiBasis.hpp will be better if #278 is merged

stefantakacs:
The last commit provides comparization operators.

I think, we should provide them using some appropriate technique (like macors as i did). Alternatives are appreciated.

I provide them also for the corners and the sides. This fixes also a very hard-to-detect bug in patchCorner, which provides its own operator== but no operator< (and friends). This does not mean not(!) mean that pc1<pc2 is not possible. No, it delegates to boxCorner which yields completely unexpected behavior!

Therefore, I suggest to provide everything needed and to explicitly delete everything that can go wrong.

src/gsCore/gsBoxTopology.cpp
```diff
@@ -210,6 +210,87 @@ int gsBoxTopology::getMaxValence() const
     return maxValence;
 }
 
+namespace {
+
+patchCorner getCanonicCorner( const patchCorner& c, const gsBoxTopology& bt )
```
filiatra: please add brief doc, even if they are hidden functions..

src/gsCore/gsMultiBasis.h
```diff
@@ -422,6 +422,38 @@ class gsMultiBasis : public gsFunctionSet<T>
...
+
+    /// @brief Returns the basis that corresponds to the component
+    ///
+    /// @param p         The component
+    /// @param dm        The dof mapper to be used
+    /// @param indices   The row vector where the indices are stored to
+    /// @param no_lower  If true, the transfer matrix does not include parts belonging to lower-order
+    ///                  components (i.e., edges without corners or faces without corners and edges)
+    typename gsBasis<T>::uPtr componentBasis_withIndices(
+        patchComponent p,
```
filiatra: patchComponent is const ?

src/gsCore/gsMultiBasis.h
```diff
@@ -422,6 +422,38 @@ class gsMultiBasis : public gsFunctionSet<T>
...
+    
+    /// @brief Returns the bases that correspond to the components
+    ///
+    /// @param p         The component
+    /// @param dm        The dof mapper to be used
+    /// @param indices   The row vector where the indices are stored to
+    /// @param no_lower  If true, the transfer matrix does not include parts belonging to lower-order
+    ///                  components (i.e., edges without corners or faces without corners and edges)
+    std::vector<typename gsBasis<T>::uPtr> componentBasis_withIndices(
+        std::vector<patchComponent> p,
```
filiatra: std::vector<patchComponent> p
should be const reference ?

src/gsCore/gsMultiBasis.hpp
```diff
@@ -249,10 +249,91 @@ void gsMultiBasis<T>::uniformCoarsen_withTransfer(
...
+        const index_t loc = indices(i,0);
+        if (dm.is_free(loc, pc.patch()))
+        {
+            indices(j,0) = dm.index(loc, pc.patch());
+            ++j;
+        }
+    }
+    // Eigen does delete all data on resize. Thus, we have to do it on
+    // our own. We consider two important special cases: j==0 and j==sz.
+    if (j==0)
```
filiatra: use:
indices.conservativeResize(j,1);
so that Eigen does not delete all data.

src/gsCore/gsBoundary.h
```diff
@@ -428,12 +407,196 @@ struct patchCorner : public boxCorner
...
+/**
+ *   @brief Struct which represents a certain component (interior, face, egde, corner).
+ *
+ *   This struct is a generalization of \a boxSide and of \a boxCorner.
+ *
+ */
+
+struct GISMO_EXPORT boxComponent {
+private:
+    index_t m_index;        ///< The index defines the component
```
filiatra: Can you add short example of the values of the m_index as a comment ?
How does it correspond in 2D ? a couple of instances in 3D ?

src/gsCore/gsBoundary.h
```diff
@@ -428,12 +407,196 @@ struct patchCorner : public boxCorner
...
+};
+
+/**
+ *   @brief Struct which represents a certain component (interior, face, egde, corner) of a particular patch
+ *
+ *   This struct is a generalization of \a patchSide and of \a patchCorner.
+ *
+ */
+
+struct GISMO_EXPORT patchComponent : boxComponent {
```
filiatra: Maybe adding m_patch as a member of boxComponent (with default value 0) and completely removing patchComponent simplifies things ? 
Or is there a real advantage to have them separately ?

(I pose it as a question but we can do this, if needed, in a new PR, so not to be changed in this PR)

stefantakacs: I have followed the design of box/patchCorner and box/patchSide.

I think an alternative design (like what you have proposed) is possible.

src/gsCore/gsMultiBasis.h
stefantakacs: It is passed by value (since it's not larger than a pointer). Thus, the caller does not need to know.

But, I can change it to const.

src/gsCore/gsMultiBasis.hpp
stefantakacs: Oh, thanks.

src/gsUtils/gsUtils.h
```diff
@@ -218,5 +218,42 @@ std::size_t hash_range(T const * start, const T * const end)
 
 } // end namespace util
 
+#define GISMO_DELEGATING_COMPARIZATION_OPERATORS( T )               \
```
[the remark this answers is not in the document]
stefantakacs: done

src/gsCore/gsBoxTopology.cpp
```diff
@@ -210,6 +210,87 @@ int gsBoxTopology::getMaxValence() const
...
+{
+    std::vector< patchCorner > corners;
+    bt.getCornerList(c,corners);
+    std::sort(corners.begin(), corners.end());
+    return corners[0];
+}
+
+std::vector<patchCorner> getCanonicCorners( const std::vector<patchCorner>& c, const gsBoxTopology& bt )
```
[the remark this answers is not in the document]
stefantakacs: done

src/gsCore/gsMultiBasis.h
stefantakacs: done

src/gsCore/gsBoundary.h
stefantakacs: done

src/gsCore/gsBoxTopology.cpp
stefantakacs: done

stefantakacs:
Error when uploading file: /Users/distiller/gismo/Testing/20190328-1046/Configure.xml
   Error message was: The requested URL returned error: 405 Not Allowed

filiatra:
This problem is related to the cdasj ricam server rejectinh the submissions. I have sent an email to Florian, but maybe you can also ask him as well..
Otherwise as we see there are no errors..

src/gsCore/gsMultiBasis.h
filiatra: Ok as you wish, I thought it is a few ints, so larger maybe than a pointer

stefantakacs:
Yes, thanks

PR #280, opened 2018-12-18
fix for https://github.com/gismo/internal/issues/4

codewerfer:
fix for https://github.com/gismo/internal/issues/4

### DIFF
--- a/examples/adaptiveConvectionDiffusionReaction_example.cpp
+++ b/examples/adaptiveConvectionDiffusionReaction_example.cpp
@@ -124,7 +124,7 @@ int main(int argc, char *argv[])
    // Construct assembler
    gsCDRAssembler<real_t> cdrAss( cdrPde, bases);
    // Set stabilization flag to 1 = SUPG
-   cdrAss.options().setInt("Stabilization", gsVisitorCDR<real_t>::SUPG);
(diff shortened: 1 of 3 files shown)


codewerfer:
I see. Nice..
But, not in `https://github.com/gismo/gismo/blob/stable/src/gsAssembler/gsAssemblerOptions.h` ?

stefantakacs:
I would not put it to gsAssemblerOptions because its only related to this one assembler

stefantakacs:
Can you confirm that the flag is used somewhere and/or confirm that it is passed to gsVisitorCDR ?

I think the class has some issues. The doxygen comments are rather wrong.

codewerfer:
> I would prefer an enum:
> 
> class gsCDRAssembler{
> 
> //...
> enum stabilizationMethod { noStabilization=0, SUPG=1 };
> 
>// ...
> 
> };
> 
> and take only parameters of that type. The gsOptionList is kept as-is.
> 
> That would be consistent with dirichletStrategy and iFaceStrategy.

> I would not put it to gsAssemblerOptions because its only related to this one assembler

OK, but to be consistent, this means also to use the enum in `gsVisitorCDR.h`, and that means an include of `gsCDRAssembler.h`, that includes already `gsVisitorCDR.h`. \
Therefore it makes more sense to add the enum to `gsVisitorCDR.h` instead.

PR #282, opened 2018-12-20
gsAssembler::updateSolution works for vector valued functions

Shamanskiy:
Why copying the visitor? This may require a copy constructor

### DIFF
--- a/src/gsAssembler/gsAssembler.h
+++ b/src/gsAssembler/gsAssembler.h
@@ -450,21 +450,23 @@ public: /* Element visitors */
     /// @brief Iterates over all elements of the domain and applies
     /// the \a ElementVisitor
     template<class ElementVisitor>
-    void push(ElementVisitor & visitor)
+    void push(const ElementVisitor & visitor)
     {
         for (index_t np = 0; np < m_pde_ptr->domain().nPatches(); ++np)
         {
+            ElementVisitor curVisitor = visitor;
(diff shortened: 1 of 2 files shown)


Shamanskiy:
I changed gsAssembler::updateSolution so that it works for vector valued functions as well, i.e. for solutions with several unknowns, each unknown has dimension one. Inspired by gsAssembler::constructSolution

src/gsAssembler/gsAssembler.h
```diff
@@ -454,9 +454,8 @@ class gsAssembler
     {
         for (index_t np = 0; np < m_pde_ptr->domain().nPatches(); ++np)
         {
-            ElementVisitor curVisitor = visitor;
             //Assemble (fill m_matrix and m_rhs) on patch np
-            apply(curVisitor, np);
+            apply(visitor, np);
```
filiatra: Hi Alex, do you need this change or is it just to avoid the extra copy ?
the copy is needed for paralellization...

Shamanskiy: No, I don't need it. I was just wondering: in the next function , push a nontrivial boundary visitor, there is no copy. Now I remember that we did it together in Delft in a hurry. I ll put a copy there then.  https://github.com/gismo/gismo/blob/bca4e24e5addc7297c7708e9b94fde1b73d3cd0f/src/gsAssembler/gsAssembler.h#L463-L469

PR #287, opened 2019-01-18
Issue3

codewerfer:
fix part of https://github.com/gismo/internal/issues/3

* change branch to master for submodules at init. (this affects all cmake . builds)
  * add KEEPBRANCH parameter - ignores master branch
  * logic to change branch in every submodule individual


codewerfer:
ok, don't get it why there is a linker error

codewerfer:
@stefantakacs I think you can squash and merge it.

filiatra:
Hi, I updated the ctest script, please adjust the cronjobs and check that all works OK,
there is a new iption -D GISMO_SUBMODULES to be set as a list of submodules to be activated.

codewerfer:
> Hi, I updated the ctest script, please adjust the cronjobs and check that all works OK,
> there is a new iption -D GISMO_SUBMODULES to be set as a list of submodules to be activated.

Fails on both Windows and Linux Continuous builds - today was the first run of the new script.

Logfile from Windows gives no information - and exited with LastTaskResult 0. But under Linux there is some information:
> CMake Error at /home/gismo/cdash/cdash/ctest_script_cont.cmake:593 (if):
  if given arguments:
    "GREATER" "0"
  Unknown arguments specified

https://github.com/gismo/gismo/blob/0ddcc29ac5a94f0a39d3b29a25ca2415fd196717/cmake/ctest_script.cmake#L592-L593
CMake is very delicate how somebody should handle a return parameter. My tests shown that it must be set with `set(${varname} ... PARENT_SCOPE)`, and no usage as like `math(EXPR varname "...")` or `math(EXPR ${varname} "...")` should be used or makes sense.

```cmake
function(myF r1 r2 r3 r4 r5 r6)
  #a
  set(r1 1 PARENT_SCOPE)
  #b
  set(${r2} 2 PARENT_SCOPE)
  #c
  set(${r3} 3 PARENT_SCOPE)
  math(EXPR r3 "4")
  #d
  set(${r4} 5 PARENT_SCOPE)
  math(EXPR ${r4} "6")
  #e
  math(EXPR r5 "7")
  set(${r5} 8 PARENT_SCOPE)
  #f
  math(EXPR ${r6} "9")
  set(${r6} 10 PARENT_SCOPE)
 
  message("r1: ${r1}")
  message("r2: ${r2}")
  message("r3: ${r3}")
  message("r4: ${r4}")
  message("r5: ${r5}")
  message("r6: ${r6}")
endfunction()

myF(a b c d e f)
message(a:${a})
message(b:${b})
message(c:${c})
message(d:${d})
message(e:${e})
message(f:${f})
```
Result:
> r1: a
r2: b
r3: 4
r4: d
r5: 7
r6: f
a:
b:2
c:3
d:5
e:
f:10

So, my best solution was to use local variables and parse the result to the parameter at the end of a function with a `set(${parameter} ${localvar} PARENT_SCOPE)`. (There was a reason why I had so many messages for everything in the code).

cmake/ctest_script.cmake
```diff
@@ -396,15 +462,37 @@ endif()
...
+    OUTPUT_VARIABLE gitHash)
+    execute_process(COMMAND ${CTEST_UPDATE_COMMAND} submodule
+      WORKING_DIRECTORY ${CTEST_SOURCE_DIRECTORY}
+      OUTPUT_VARIABLE submoduleHashes)
+    set(${res} " ${gitHash} gismo\n${submoduleHashes}\n")
+endmacro(get_git_status)
+
+function(update_gismo ug_ucount)
+  #this should be last.
+  ctest_update(SOURCE ${CTEST_SOURCE_DIRECTORY} RETURN_VALUE updcount)
```
codewerfer: Think this updates submodules too - if the change on HEAD, will be get the update count?

filiatra: This should update the submodules not to HEAD, but detached mode..

filiatra:
I think the return argument should work now (did not check yet!).

cmake/ctest_script.cmake
```diff
@@ -482,7 +483,7 @@ function(update_gismo ug_ucount)
 	WORKING_DIRECTORY ${CTEST_SOURCE_DIRECTORY}/extensions/${submodule})
       ctest_update(SOURCE ${CTEST_SOURCE_DIRECTORY}/extensions/${submodule} RETURN_VALUE ug_upd_sm)
       if (${ug_upd_sm} GREATER 0)
-	math(EXPR updcount "${updcount} + ${ug_upd_sm}")
+	math(EXPR ${ug_ucount} "${${ug_ucount}} + ${ug_upd_sm}")
```
codewerfer: You have no write access to the variable/parameter at **PARENT_SCOPE** via `math(EXPR ${ug_ucount} "...")`.
Only `set(${ug_ucount} ... PARENT_SCOPE)` has write access to the variable/parameter at parent scope.
(And only if it is a parameter (a new variable) at parent scope, otherwise it seams we have no chance to write to it, but well, we do not want to override variables of the PARENT_SCOPE by accident, like using the same name twice).

Think of it like 2 separate variables, one is a copy you can access with `${var}`, the other is a reference you can't access normally, but the `set(${var} ... PARENT_SCOPE)` can - the `set(var ...)` will break it up. So every write to `${var}` modifies only the copy. And nearly every read of `${var}` is only one of the copy. Nearly, because a `set(${var} ${var} PARENT_SCOPE)` has the reference for all parameters that are not defined local (with `set(var ...)`) on both sides and doesn't do anything in our case. - Well, because our ${var} gets to be a variable with the name par (if function f was called as f(par)) it's getting a bit complicate and we should look not to use names twice, write isn't the problem there, it's the read, that should be the parent before a write and the local after a write and is a bit brain dizzying. 

Therefore on the local scope it seams to work - I think you tested it with the out-commented `#if (${ug_ucount} GREATER 0) endif()`, but on the parent scope, it's not known that something changed - or that the variable `updcount` has a value. So we result in an error saying **`if( GREATER 0)` doesn't work**.

By the way, you do not include the result value of `ctest_update(SOURCE ${CTEST_SOURCE_DIRECTORY} RETURN_VALUE updcount)`. OK, seams you thought that it is the same parameter(name), but not the same scope.

filiatra: Probably macros work more the way we want, so I changed to a macro now!

codewerfer: Macros do also need the PARENT_SCOPE if I remember correct to get out. All variables are just copies. My girlfriend say's I should stop now - I'm still at home and **this** keeps me busy.

cmake/ctest_script.cmake
```diff
@@ -426,9 +514,9 @@ macro(run_ctests)
   endif()
   #set(CTEST_LABELS_FOR_SUBPROJECTS ${LABELS_FOR_SUBPROJECTS}) #!Dangerous!
 
-  ctest_configure(OPTIONS "${CMAKE_ARGS};-DCTEST_USE_LAUNCHERS=${CTEST_USE_LAUNCHERS};-DBUILD_TESTING=ON;-DDART_TESTING_TIMEOUT=${CTEST_TEST_TIMEOUT}")
+  ctest_configure(OPTIONS "${CMAKE_ARGS};${SUBM_ARGS};-DCTEST_USE_LAUNCHERS=${CTEST_USE_LAUNCHERS};-DBUILD_TESTING=ON;-DDART_TESTING_TIMEOUT=${CTEST_TEST_TIMEOUT}")
 
-  ctest_submit(PARTS Configure Update)
+  ctest_submit(PARTS Configure Update Notes)
```
codewerfer: This happens immediately, or?

filiatra: I did not think of this, will make a new commit to make sure that it is the case.

PR #290, opened 2019-02-26
Transposing PK_BSURF in gsWriteParasolid.

dominik-mokris-mtu:
The function `createPK_BSURF` from extensions/gsParasolid/gsWriteParasolid swaps _u_- and _v_- directions (the 0-th parametric dimension of Gismo becomes _v_ and the 1-st becomes _u_ in Parasolid). This is caused by Parasolid and Gismo storing the control points in different order (row-wise vs. col-wise).

While the geometry remains unchanged, it is still a bug, as we have to transpose after each export. I have committed a simple solution: we create a `PK_BSURF` with the directions swapped and then transpose the geometry with `PK_BSURF_reparameterise`.

We are the sole users of the Parasolid interface, so I did not add any further tests to Gismo.

### DIFF
--- a/extensions/gsParasolid/gsWriteParasolid.hpp
+++ b/extensions/gsParasolid/gsWriteParasolid.hpp
@@ -312,6 +312,8 @@ bool createPK_BSURF(const gsTensorBSpline< 2, T> & bsp,
                     bool closed_u,
                     bool closed_v)
 {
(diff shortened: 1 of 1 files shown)


dominik-mokris-mtu:
Thanks. Would it be possible that also @filiatra would have a quick look? I think it is originally written by him.

Issue #292, opened 2019-03-19
Linker errors in gsElasticity


Shamanskiy:
I checked out the latest version of stable, and the linker error appeared when trying to compile gsElasticity/examples. 
The library itself compiles, the error appears only when compiling executables:
![Screenshot from 2019-03-19 13-42-44](https://user-images.githubusercontent.com/29507068/54606801-12cef100-4a4d-11e9-8624-0ad8399a307c.png)

May it be related to some recent changes in CMake files?

P.S. If you want to check it, don't forget to fetch the latest version of gsElasticity manually, it is not yet in gismo-stable

codewerfer:
**Yes**, there is something wrong with the includes - did not know at the moment why exactly it is.
(gsElasticityAssembler{.h,.hpp,_.cpp} exists)

For the `gsElasticityAssembler` stuff, it works - but this is not the fix! - if you change [line 3](https://github.com/gismo/gsElasticity/blob/master/examples/linElast2D_lshape.cpp#L3) to:
```c++
#include <gsElasticity/gsElasticityAssembler.hpp>
```

Keeps us up with `linElast2D_lshape.cpp:103: undefined reference to `void gismo::gsWriteParaviewMultiPhysics<double>(`
There the same include stuff.

If we grep for the build object file - and compare the result with one where we know it works, we see that it is missing in some `build/CMakeFiles/<???>/link.txt` file.

> May it be related to some recent changes in CMake files?

I think also so.
Btw. to get a better view in github to compare latest with known working hash, use this link:
https://github.com/gismo/gsElasticity/compare/78a79ae00e5b8551275a2c6f7b278b129b660357...master

PR #295, opened 2019-03-27
Fix gsElastity linker error

Shamanskiy:
This fixes the linker error #292 where gsElasticity examples can't be compiled. I don't know why, but apparently https://github.com/gismo/gismo/blob/8f2e332e33249cf6275327f0bb97b64daffceff4/CMakeLists.txt#L393
has to come before https://github.com/gismo/gismo/blob/8f2e332e33249cf6275327f0bb97b64daffceff4/CMakeLists.txt#L313
as it used to be before the 8a6c6d0fbc02d733165e20a8bc0291f6f7da37b6. It probably has to be the same for the rest of the submodules, but I didn't touch it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -310,6 +310,10 @@ if(GISMO_BUILD_LIB)
   endif()
 endif()
 
+if(GISMO_ELASTICITY)
+  gismo_fetch_module(gsElasticity)
+endif()
+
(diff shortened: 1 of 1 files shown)


filiatra:
Thnx, Can you please also delete gsElasticity from unsuported ?

Shamanskiy:
Ok, I will do it

PR #296, opened 2019-03-28
Update gsElasticity submodule

Shamanskiy:
Set gismo to track the latest version of gsElasticity.

Can we also make gsElasticity public now? It is ready


filiatra:
Does it contain everything from unsupported ?
In that case, shall we first remove the old files ?
Then we can proceed towards making public ..

Shamanskiy:
I have already removed gsElasticity from the unsupported submodule.

filiatra:
ok that's great, I wll have a last look at gsElasticity and then make it public

PR #297, opened 2019-03-29
Dim t

codewerfer:
Introduced `dim_t` as type for dimension.

`dim_t` is set in `gsConfig.h.in` (*for testing purpose it is **short** now, will be set to **int** before merge*)

related Pull request will exist in unsupported and motor.


codewerfer:
missing: https://github.com/gismo/gismo/compare/3cd059bfefb58901c59976babcb927f087fb7f01...stable

codewerfer:
OK, THX
Will first merge the 3 subprojects and update them back into this one.

PR #301, opened 2019-04-15
add unsigned T abs_diff(T a, T b)

codewerfer:
On some code (unsupported, like [gsIETIdG_experimental.cpp#L145](https://github.com/gismo/unsupported/blob/b36880528353bac9826de7892fe4338679e2032e/examples/gsIETIdG_experimental.cpp#L145)) there is abs(a - b) used to find out the distance of 2 indexes.
Since this will not work with size_t, and is buggy in special cases, it would make sense to add an error free implementation that works well with all integer types.

### DIFF
--- a/src/gsCore/gsMath.h
+++ b/src/gsCore/gsMath.h
@@ -326,7 +326,8 @@ template <typename T> int getSign(T val)
 template <typename T>
 inline typename util::make_unsigned<T>::type abs_diff(T a, T b)
 {
-    return a > b ? a - b : b - a;
+    typedef typename util::make_unsigned<T>::type unsignedT;
+    return a > b ? unsignedT(a) - unsignedT(b) : unsignedT(b) - unsignedT(a);
 }
 
 /// integer power
--- a/src/gsCore/gsMath.h
+++ b/src/gsCore/gsMath.h
@@ -324,7 +324,7 @@ template <typename T> int getSign(T val)
 /// Return smallest difference between two (integer) numbers as size_t
 /// especially the case abs_diff(INT32_MAX, INT32_MIN) := UINT32_MAX is correct
 template <typename T>
-inline typename std::make_unsigned<T>::type abs_diff(T a, T b)
+inline typename util::make_unsigned<T>::type abs_diff(T a, T b)
 {
     return a > b ? a - b : b - a;
 }
--- a/src/gsCore/gsMath.h
+++ b/src/gsCore/gsMath.h
@@ -321,6 +321,18 @@ template <typename T> int getSign(T val)
     return (T(0) < val) - (val < T(0));
 }
 
+/// Return smallest difference between two (integer) numbers as size_t
+/// especially the case abs_diff(INT32_MAX, INT32_MIN) := UINT32_MAX is correct
+template <typename T>
+inline size_t abs_diff(T a, T b)
+{
+    if (getSign(a) != getSign(b))
+    {
+        return (size_t) math::max(a, b) + -(math::min(a, b) + 1) + 1; // false for abs_diff(UINT32_MAX, UINT32_MAX)
+    }
+    return a > b ? a - b : b - a; // false for abs_diff(>=0,INT32_MIN)
+}
+
 /// integer power
 inline int ipow(int x, unsigned exp)
 {
--- a/src/gsCore/gsTemplateTools.h
+++ b/src/gsCore/gsTemplateTools.h
@@ -138,7 +138,7 @@ template<typename T> struct is_integral: is_integral_base<typename remove_cv<T>:
 template<class T>
 struct make_unsigned;
 
-#define MAKE_UNSIGNED(signed_type)           \
+#define GISMO_MAKE_UNSIGNED(signed_type)           \
 template<>                                   \
 struct make_unsigned<signed_type> {          \
     typedef unsigned signed_type type;       \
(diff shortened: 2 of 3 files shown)


src/gsCore/gsMath.h
```diff
@@ -321,6 +321,18 @@ template <typename T> int getSign(T val)
     return (T(0) < val) - (val < T(0));
 }
 
+/// Return smallest difference between two (integer) numbers as size_t
+/// especially the case abs_diff(INT32_MAX, INT32_MIN) := UINT32_MAX is correct
+template <typename T>
+inline size_t abs_diff(T a, T b)
```
stefantakacs: Why size_t?

It should be the unsigned type fitting to T (cf. https://en.cppreference.com/w/cpp/types/make_unsigned)

src/gsCore/gsMath.h
```diff
@@ -321,6 +321,18 @@ template <typename T> int getSign(T val)
...
+{
+    if (getSign(a) != getSign(b))
+    {
+        return (size_t) math::max(a, b) + -(math::min(a, b) + 1) + 1; // false for abs_diff(UINT32_MAX, UINT32_MAX)
+    }
+    return a > b ? a - b : b - a; // false for abs_diff(>=0,INT32_MIN)
```
stefantakacs: what means "false for"?

src/gsCore/gsMath.h
codewerfer: THX, didn't know that.

codewerfer: cool, that needs not the code to prevent abs_diff(>=0,INT32_MIN)

stefantakacs: its c++11 only. so, we need our own version

stefantakacs: like in tensortools

template<class T> struct make_unsigned {};
template<> struct<int> make_unsigned { typedef unsigned type; };
// ...

src/gsCore/gsMath.h
```diff
@@ -321,6 +321,14 @@ template <typename T> int getSign(T val)
     return (T(0) < val) - (val < T(0));
 }
 
+/// Return smallest difference between two (integer) numbers as size_t
+/// especially the case abs_diff(INT32_MAX, INT32_MIN) := UINT32_MAX is correct
+template <typename T>
+inline typename std::make_unsigned<T>::type abs_diff(T a, T b)
+{
+    return a > b ? a - b : b - a;
```
stefantakacs: i think, one has to cast a and b first to the resulting type

a > b ? static_cast<...>(a) - static_cast<...>(b) : static_cast<...>(b) - static_cast<...>(a)

stefantakacs: to avoid overflow!?

src/gsCore/gsMath.h
codewerfer: > what means "false for"?
Filling up the leading 1 at cast to size_t. Therefore "18446744071562067968" instead of "2147483648".

src/gsCore/gsMath.h
codewerfer: > like in tensortools
> 
> template<class T> struct make_unsigned {};
> template<> struct make_unsigned<int> { typedef unsigned type; };
> template<> struct make_unsigned<unsigned> { typedef unsigned type; };
> // ...

can't find it in our code. But like [this](https://stackoverflow.com/questions/18183919/c-template-integer-types-adequate-unsigned-type)?

src/gsCore/gsMath.h
stefantakacs: Yes, https://timsong-cpp.github.io/cppwp/n4659/conv.integral allows it

src/gsCore/gsTemplateTools.h
```diff
@@ -134,6 +135,24 @@ struct remove_cv { typedef typename remove_volatile<typename remove_const<T>::ty
...
+template<>                                   \
+struct make_unsigned(unsigned signed_type) { \
+    typedef unsigned signed_type type;       \
+};
+
+MAKE_UNSIGNED(short)
+MAKE_UNSIGNED(int)
+MAKE_UNSIGNED(long)
+MAKE_UNSIGNED(long long)
+
```
stefantakacs: better to rename to GISMO_MAKE_UNSIGNED to avoid name clashes.

maybe also good to undefine the macro afterwards

src/gsCore/gsTemplateTools.h
```diff
@@ -134,6 +135,24 @@ struct remove_cv { typedef typename remove_volatile<typename remove_const<T>::ty
...
+template<class T>
+struct make_unsigned;
+
+#define MAKE_UNSIGNED(signed_type)           \
+template<>                                   \
+struct make_unsigned<signed_type> {          \
+    typedef unsigned signed_type type;       \
+};                                           \
+template<>                                   \
+struct make_unsigned(unsigned signed_type) { \
```
stefantakacs: &lt; and not (

src/gsCore/gsTemplateTools.h
codewerfer: ok, will rename.
Btw. a Unittest is also already in unsupported: https://github.com/gismo/unsupported/blob/master/unittests/gsMakeUnsigned.cpp

src/gsCore/gsMath.h
```diff
@@ -326,7 +326,8 @@ template <typename T> int getSign(T val)
 template <typename T>
 inline typename util::make_unsigned<T>::type abs_diff(T a, T b)
 {
-    return a > b ? a - b : b - a;
+    typedef typename util::make_unsigned<T>::type unsignedT;
```
codewerfer: not sure if this is really needed.
The Unittest runs without very fine.

stefantakacs: but it is ub

codewerfer: ub?

PR #302, opened 2019-04-16
fix a small bug in gsBoxTopology.cpp

stefantakacs:


src/gsCore/gsBoxTopology.cpp
```diff
@@ -292,8 +292,9 @@ std::vector< std::vector<patchComponent> > gsBoxTopology::allComponents(bool com
             {
                 const index_t nrcp = it->second.size();
                 for (index_t j=0; j<nrcp; ++j)
-                    last.push_back(it->second[i]);
+                    last.push_back(it->second[j]);
```
filiatra: this is inefficient, use 
std::vector::insert (iterator position, size_type n, const value_type& val);
instead

PR #309, opened 2019-05-02
Prevent misuses of gsCmdLine

stefantakacs:


stefantakacs:
Let's see how many tests it breaks...

codewerfer:
Is one of this builds here not a release?

stefantakacs:
I don't know. I am checking locally if there are tests that should be broken.

codewerfer:
OK

stefantakacs:
I do more testing before merging

PR #315, opened 2019-05-09
fix -Wsign-compare warning

codewerfer:
this should fix https://cdash.ricam.oeaw.ac.at/viewBuildError.php?type=1&buildid=337199

### DIFF
--- a/src/gsCore/gsTemplateTools.h
+++ b/src/gsCore/gsTemplateTools.h
@@ -249,13 +249,13 @@ bool less(T1 t1, T2 t2)
     {
         if (t1 < 0)
             return true;
-        return (static_cast<unsignedT1>(t1) < t2);
+        return (static_cast<unsignedT1>(t1) < static_cast<unsignedT2>(t2));
     }
     if (!is_signed<T1>::value && is_signed<T2>::value)
     {
         if (t2 < 0)
             return false;
-        return (t1 < static_cast<unsignedT2>(t2));
+        return (static_cast<unsignedT1>(t1) < static_cast<unsignedT2>(t2));
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/gsCore/gsTemplateTools.h
```diff
@@ -233,13 +233,25 @@ typename make_signed<T>::type to_signed(T t) {
...
+            return (unsignedT1(t1) < unsignedT2(t2));
+    }
+
     if (t1 < 0 && t2 >= 0)
         return true;
     else if (t2 < 0 && t1 >= 0)
         return false;
-    return (static_cast<size_t>(t1) < static_cast<size_t>(t2));
+
+    return (static_cast<size_t>(t1) < static_cast<size_t>(t2)); // both positive and fit in size_t
```
stefantakacs: Why size_t?

src/gsCore/gsTemplateTools.h
```diff
@@ -233,13 +233,25 @@ typename make_signed<T>::type to_signed(T t) {
...
-        return (t1 < t2);
+    typedef typename util::make_signed<T1>::type signedT1;
+    typedef typename util::make_signed<T2>::type signedT2;
+    typedef typename util::make_unsigned<T1>::type unsignedT1;
+    typedef typename util::make_unsigned<T2>::type unsignedT2;
+
+    if (is_signed<T1>::value == is_signed<T2>::value) // all is_signed are optimized out at compile time
+    {
+        if (is_signed<T1>::value) // both signed, cast to signedTx
+            return (signedT1(t1) < signedT2(t2));
```
stefantakacs: Here, you are casting to T1 and T2, i.e., you are not casting.

src/gsCore/gsTemplateTools.h
stefantakacs: We said that ut should be cased to unsignedT1 and unsigned T2.

src/gsCore/gsTemplateTools.h
codewerfer: I'm not happy with the case:
```cpp
for (int i = 0; less(i, 10ull; ++i) {
    ...
} 
```
It looks like on godbolt that the compiler can't optimize it out - knows that i is positive - therefore goes trough the branches.

Btw. I read [Es.100 - 107](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es107-dont-use-unsigned-for-subscripts-prefer-gslindex) and now I do not know what would be best. Can we meet up?

PR #316, opened 2019-05-10
Vs warnings [NOT MERGE]

codewerfer:
need the cdash log for VS Warnings


filiatra:
Please fix warnings
warns: https://cdash.ricam.oeaw.ac.at/viewBuildError.php?type=1&buildid=350477

codewerfer:
> Please fix warnings
> warns: https://cdash.ricam.oeaw.ac.at/viewBuildError.php?type=1&buildid=350477

This affects `gsBasis::active_into` too. Therefore I will start the `gsBasis::active_into` task as you told me on our last meeting first. Btw. the `-Wconversion` flag works to find that warnings under Linux.

filiatra:
OK, in any case you could finish this PR first by fixing any warnings that do not depend on active_into type..

codewerfer:
I tried, but I run into active_into with size_t instead of index_t like you want it. Fixing first this Warnings and refactor active_into to index_t would result in many new warnings just because I swaped the data types. I suggest to first set the active_into to the right data type, to have an anchor from where I can see what datatype should be the right one for the rest - and fix than the warnings based on this anchor.

filiatra:
ok sure!

PR #319, opened 2019-05-22
simplify gsPatchPreconditionersCreator

stefantakacs:
We can directly use the virtual function gsBasis<T>::component() instead of casting to gsTensorBSplineBasis first. Only one function really relies on B-Splines.

### DIFF
--- a/src/gsSolver/gsPatchPreconditionersCreator.hpp
+++ b/src/gsSolver/gsPatchPreconditionersCreator.hpp
@@ -389,8 +389,8 @@ template<typename T>
 void tildeSpaceBasis(const gsBasis<T>& basis, gsSparseMatrix<T>& B_tilde, gsSparseMatrix<T>& B_compl, const gsBoundaryConditions<T>& bc, const bool odd = true)
 {
     const gsTensorBSplineBasis<1,T>* bbasis_ptr = dynamic_cast<const gsTensorBSplineBasis<1,T>*>(&basis);
-    GISMO_ENSURE( bbasis_ptr, "gsPatchPreconditionersCreator<T>::getTildeSpaceBasisTransformation and "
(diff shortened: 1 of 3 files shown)


src/gsSolver/gsPatchPreconditionersCreator.hpp
```diff
@@ -424,8 +386,13 @@ void tildeSpaceBasis_oneside(const gsTensorBSplineBasis<1,T>& basis, bool isLeft
 
 // Compute a basis for S-tilde and one for its orthogonal complement
 template<typename T>
-void tildeSpaceBasis(const gsTensorBSplineBasis<1,T>& basis, gsSparseMatrix<T>& B_tilde, gsSparseMatrix<T>& B_compl, const gsBoundaryConditions<T>& bc, const bool odd = true)
+void tildeSpaceBasis(const gsBasis<T>& basis, gsSparseMatrix<T>& B_tilde, gsSparseMatrix<T>& B_compl, const gsBoundaryConditions<T>& bc, const bool odd = true)
 {
+    const gsTensorBSplineBasis<1,T>* bbasis_ptr = dynamic_cast<const gsTensorBSplineBasis<1,T>*>(&basis);
+    GISMO_ENSURE( bbasis_ptr, "gsPatchPreconditionersCreator<T>::getTildeSpaceBasisTransformation and "
```
filiatra: Compare pointers to nullptr to avoid warnings.

-This could be done even alternatively as follows:

GISMO_ASSERT(nullptr!=dynamic_cast<const gsTensorBSplineBasis<1,T>*>&basis),"..")
const gsTensorBSplineBasis<1,T>& bbasis = static_cast<gsTensorBSplineBasis<1,T>&>(basis);

the error  will only appear in debug mode. In all cases,

PR #320, opened 2019-05-23
More functionality in gsFileManager

codewerfer:
Create draft pull request from https://github.com/gismo/gismo/pull/311
-----

    Allow to determine the path where the executable is located in (we could locate a config file there in future)
    Allow to provide a canonical representation of a path
    Provide better specs of all functions in gsFileManager
    Make sure that the functions behave as specified
    Unittests
    Don't expose local stuff in header file

TODO:
* [x] remove `argv0`, see https://github.com/gismo/gismo/pull/320#issuecomment-495636174
* [x] Unittest for `getHomePath`
* [x] Unittest for `makeRelative`
* [x] Check all under Windows
    * [x] fix Unittests (Path Separator)

Because we have a `fileExists` and a `mkdir`, a `dirExists` would make some sense.

### DIFF
--- a/src/gsIO/gsFileManager.cpp
+++ b/src/gsIO/gsFileManager.cpp
@@ -280,7 +280,11 @@ std::string gsFileManager::find(std::string fn)
     {
         tmp = (*it) + fn;
         if ( _fileExistsWithoutSearching(tmp) )
+        {
+            if(isExplicitlyRelative(tmp))
+                tmp = getCanonicRepresentation(getCurrentPath() + tmp);
             return tmp;
+        }
     }
 
     return std::string();
@@ -432,6 +436,33 @@ std::string gsFileManager::getHomePath()
     std::string path(_temp);
     _makePath(path);
     return path;
+}
+
+std::string gsFileManager::makeRelative(const std::string & from, const std::string & to)
+{
+    if (!isFullyQualified(from))
+        return "";
+
+    std::string fromc = getCanonicRepresentation(from);
+    std::string toc = getCanonicRepresentation((isFullyQualified(to) ?
+        "": getCurrentPath()) + to);
+
+    size_t start = 0;
+    size_t pos = 0;
+    while(fromc[pos] == toc[pos]){
+        if(fromc[pos++] == getNativePathSeparator())
+            start = pos;
+    }
+
+    std:: string result;
+    for (size_t i = 0; i < util::count(fromc.substr(start), getNativePathSeparator()); ++i)
+    {
+        result += "..";
+        result.push_back(getNativePathSeparator());
+    }
+    result += toc.substr(start);
+
+    return result;
 }
 
 bool gsFileManager::pathEqual( const std::string& p1o, const std::string& p2o )
--- a/src/gsIO/gsFileManager.cpp
+++ b/src/gsIO/gsFileManager.cpp
@@ -323,8 +323,8 @@ std::string gsFileManager::getExePath()
     }
 
     return isFullyQualified( argv0 )
-        ? getCanonicRepresentation( argv0 )
-        : getCanonicRepresentation( getCurrentPath() + "/" + argv0 );
+        ? getCanonicRepresentation( std::string(argv0) + "/../" )
+        : getCanonicRepresentation( getCurrentPath() + "/" + argv0 + "/../" );
 }
 
 bool gsFileManager::pathEqual( const std::string& p1, const std::string& p2 )
--- a/src/gsIO/gsFileManager.cpp
+++ b/src/gsIO/gsFileManager.cpp
@@ -322,7 +322,9 @@ std::string gsFileManager::getExePath()
         return getCurrentPath();
     }
 
-    return getCanonicRepresentation( getCurrentPath() + "/" + argv0 );
+    return isFullyQualified( argv0 )
+        ? getCanonicRepresentation( argv0 )
+        : getCanonicRepresentation( getCurrentPath() + "/" + argv0 );
 }
 
 bool gsFileManager::pathEqual( const std::string& p1, const std::string& p2 )
--- a/src/gsIO/gsFileManager.cpp
+++ b/src/gsIO/gsFileManager.cpp
@@ -40,18 +40,12 @@ namespace gismo
 struct GISMO_EXPORT gsFileManagerData {
     gsFileManagerData();
     std::vector<std::string> m_paths;
-    const char* argv0;
 };
 
 gsFileManagerData& gsFileManagerDataSingleton()
 {
     static gsFileManagerData singleton;
     return singleton;
-}
-
-void gsFileManager::setArgv0(const char* argv0)
-{
-    gsFileManagerDataSingleton().argv0 = argv0;
 }
 
 bool gsFileManager::fileExists(const std::string& name)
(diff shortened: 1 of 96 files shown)


filiatra:
I don't like having the name of the executable in the file manager. It is not needed andalso the information is not related to the library. I t could also be that another library is calling gismo, so there is no executable (and no name).

codewerfer:
> ... the name of the executable ...

We use argv[0] to get the the path of the executable - under Unix.

filiatra:
That is not so good, you should use
https://linux.die.net/man/3/getcwd

unittests/gsFileManager_test.cpp
```diff
@@ -0,0 +1,473 @@
...
+    CHECK(gsFileManager::pathEqual("/foo/bar/", "/foo/bar"));
+    CHECK(!gsFileManager::pathEqual("/foo/", "/bar"));
+    CHECK(!gsFileManager::pathEqual("/foo/bar/", "/bar/foo"));
+    CHECK(gsFileManager::pathEqual("/foo/./bar/", "/foo/bar"));
+    CHECK(gsFileManager::pathEqual("/foo/baz/../bar/", "/foo/bar"));
+
+    // relative
+    CHECK(gsFileManager::pathEqual("", ""));
+    CHECK(gsFileManager::pathEqual("", "./"));
+    CHECK(gsFileManager::pathEqual("", gsFileManager::getCurrentPath()));
```
codewerfer: What did it thought at that point? Can't remember.

filiatra:
I do not agree to use argv[0] for finding the path. it is not a good approach, and it is not guarranteed to work.
I think that the correct approach is "SOLUTION 5" in this post:
https://www.wikitechy.com/tutorials/linux/finding-current-executables-path-without-proc-self-exe

I strongly suggest to change the approach and use that code instead.

filiatra:
It would also be nice to have a function  homeDir() which returns the homedir of the user.
I had this implemented somewhere but I probably missed to commit it..

codewerfer:
@filiatra will implement both of your suggestions.

Question that is still open for me. How should find handle folders? Mean, assert `/foo/bar` is a folder, what should `gsFileManager::find("/foo/bar")` and `gsFileManager::find("/foo/bar/")` return?
Same for `gsFileManager::fileExists` - isn't everything a file under Linux?

filiatra:
I think we want it to find files only, as also described in the description..

src/gsUtils/gsUtils.h
```diff
@@ -138,6 +138,36 @@ inline void string_replace(std::string& str,
     }
 }
 
+/// \brief Splits a string \a str into substrings with \a ch as separator.
```
stefantakacs: You should mention that empty "items" are ignored

src/gsUtils/gsUtils.h
```diff
@@ -138,6 +138,36 @@ inline void string_replace(std::string& str,
...
+            str.erase(0, 1);
+            continue;
+        }
+        result.push_back(str.substr(0, pos));
+        str.erase(0, pos + 1);
+    }
+    if(str.size())
+        result.push_back(str);
+    return result;
+}
```
stefantakacs: erease is an expensive operation since it requires a copy of the remainder of the string. Thus, this algorithm is O(N^2).
```

{
    std::vector<std::string> result;
    size_t begin = 0;
    size_t end;
    while ((end = str.find(ch, begin)) != std::string::npos)
    {
        if (end == begin)
        {
            ++begin;
            continue;
        }
        result.push_back(str.substr(begin, end-begin));
        begin = end + 1;
    }
    end = str.size();
    if (begin<end)
        result.push_back(str.substr(begin, end-begin));
    return result;
}
```

I this should be correct (needs testing)
For sure, its only O(N).

src/gsUtils/gsUtils.h
codewerfer: Yep, used this as helperfunction for [`gsFileManager::makeRelative`](https://github.com/gismo/gismo/blob/dd34709aeff7d771093ccd7bdd6ef29bb876392e/src/gsIO/gsFileManager.cpp#L469). Therefore I ignored empty ones complete - there exists no empty subpaths, neither under Windows nor under Unix. But as you can see here: https://github.com/gismo/gismo/blob/dd34709aeff7d771093ccd7bdd6ef29bb876392e/src/gsIO/gsFileManager.cpp#L485-L508
I implemented a O(n) way without splitting and didn't thought more an this helper function - just kept it in if somebody else could need it - we can throw it away if you want.

src/gsIO/gsFileManager.h
```diff
@@ -18,93 +18,171 @@
...
-    static char getNativePathSeparator();
+    /// Get preferred native path seperator
+    static const char getNativePathSeparator();
 
-    /// Checks if the path is fully qualified
-    /// If a name starts with "/", it is considered fully qualified
+    /// Get valid path seperators
+    static const std::string& getValidPathSeparators();
+
+    /// Get system depending not allowed characters for paths and filenames
```
filiatra: rewrite as:
/// Get system-dependent invalid characters for paths and filenames

src/gsUtils/gsUtils.h
```diff
@@ -138,6 +138,36 @@ inline void string_replace(std::string& str,
...
+
+/// \brief Count occurrences of \a ch in string \a str.
```
filiatra: Please remove this function, use instead:
std::count(s0.begin(), s0.end(), char);

src/gsIO/gsFileManager.cpp
```diff
@@ -70,96 +57,157 @@ bool gsFileManager::fileExists(const std::string& name)
...
+         (fn[0] == '.' && fn[1] == '.' && fn[2] == getValidPathSeparators()[i]);
+    }
+    valid = valid && _isValidPathOrName(fn);
+    return valid;
 }
 
-bool gsFileManagerData::addSearchPaths(const std::string& paths)
+namespace {
+
+inline void _repacle_with_native_seperator(std::string& str)
```
filiatra: rename as:
_replacle_with_native_seperator

src/gsIO/gsFileManager.cpp
```diff
@@ -21,41 +21,28 @@
...
-    {
-#ifdef GISMO_SEARCH_PATHS
-        (void)addSearchPaths("" GISMO_SEARCH_PATHS);
-#endif
-    }
-
-    gsFileManagerData(const gsFileManagerData&);
-    gsFileManagerData& operator= (const gsFileManagerData&);
+// Struct for storing data
+struct GISMO_EXPORT gsFileManagerData {
```
filiatra: Do not  export the struct (remove GISMO_EXPORT)

src/gsIO/gsFileManager.h
codewerfer: done

src/gsIO/gsFileManager.cpp
codewerfer: `_replace_with_native_separator` - days of typos

src/gsUtils/gsUtils.h
codewerfer: Done.
Also as we discussed yesterday.
`template<typename T> inline T convert_to(const std::string&);` should work too - hope so.

codewerfer:
Unittests for `util::split` will be added to https://github.com/gismo/gismo/tree/Unittests

src/gsUtils/gsUtils.h
```diff
@@ -138,6 +138,46 @@ inline void string_replace(std::string& str,
...
+/// \example
+/// std::set<index_t> s split<std::set<index_t> >("1,3,2,4,3,2,1", ','); s => {1,2,3,4}
+/// std::vector<std::string> v split<std::vector<std::string> >("ab.ba.aa", '.'); v => {"ab","ba","aa"}
+/// \tparam T STL container type, like std::vector<std::string> >, but not std::map
+/// \param str input string
+/// \param ch separator
+/// \param empty if true, adds also empty parts. default is false. Makes only sense with std::string.
+/// \return a container of T<T::value_type>
+template <class T>
+inline T split(const std::string& str, char ch, bool empty = false)
```
filiatra: Where is this used ? If it is not used too much then I do not really like it.
This function should be called "tokenize" as it is the standard name.
Templating with respect to the container type is not OK. It should be done by templating over an iterator,  and use e.g. back_inserter (http://www.cplusplus.com/reference/iterator/back_inserter/)

It would be much better to always return an std::list<T> . 
Please template it over the T = value type instead, 

The convert_to functions are also a little tricky and conflict with other things in G+Smo, for instance the gsGetReal in gsIO?gsXml.h

codewerfer: > Where is this used ? If it is not used too much then I do not really like it.

At the moment it is not used - I used it in an earlier version of `gsFileManager::makeRelative`. But I thought maybe somebody else could need it.

> This function should be called "tokenize" as it is the standard name.

Oh, well, it's called `split` in [C#](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-4.8) and [Java](https://docs.oracle.com/javase/9/docs/api/java/lang/String.html#split-java.lang.String-). Therefore I thought first on this name.

> Templating with respect to the container type is not OK.

That was the whole thing on it.

> It should be done by templating over an iterator, and use e.g. back_inserter (http://www.cplusplus.com/reference/iterator/back_inserter/)
> 
> It would be much better to always return an std::list .
> Please template it over the T = value type instead,
> 
> The convert_to functions are also a little tricky and conflict with other things in G+Smo, for instance the gsGetReal in gsIO?gsXml.h

So, I throw it out!?

filiatra: If we don't need it, I would say take it out for now..

PR #321, opened 2019-05-23
std::size

codewerfer:
introduce `std::size`

### DIFF
--- a/src/gsUtils/gsCombinatorics.h
+++ b/src/gsUtils/gsCombinatorics.h
@@ -28,8 +28,8 @@ namespace gismo
 // could also be in Utils, but doxygen allows only one group for free functions
 inline unsigned factorial( unsigned n)
 {
-    GISMO_ASSERT(n<13, "Overflow when computing factorial.");
     static const unsigned precomputed[]= {1, 1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800, 39916800, 479001600};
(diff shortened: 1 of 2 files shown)


filiatra:
Is there a use case ?

stefantakacs:
pr #314 shows a use case.

There, we could/should write
```
-    GISMO_ASSERT(n<13, "Overflow when computing factorial.");
    static const unsigned precomputed[]= {1, 1, 2, 6, 24, 120, 720, 5040, 40320, 362880, 3628800, 39916800, 479001600};
+    GISMO_ASSERT(n<util::size(precomputed), "Overflow when computing factorial.");
```

filiatra:
But in this case util::size(precomputed) is equal to 13 at compile time, why not just use 13 as it is now ?

codewerfer:
> But in this case util::size(precomputed) is equal to 13 at compile time, why not just use 13 as it is now ?
> the size of the array here is unlikely to change, otherwise there will be unsigned overflow..
> In any case, it is not bad to have the util::size if somebody needs it.

Cause this probably was the case why #314 happened. `std::size` does for us - at compile time - the correct count. We can miscount - who knows how many more miscounts exists.
Next task for me on this Draft is to replace fixed values of arraycounts with `util::size`.

filiatra:
No that is wrong. The functions you added are not compile-time. It is very easy to check that. Just print something from inside the functions and you will see that it appears at **runtime**.

Only sizeof(.) is compile time.

PR #328, opened 2019-06-26
use _MSVC_LANG, as MSVC doesn't set _cplusplus to correct value

codewerfer:
see https://github.com/gismo/internal/issues/10

fixes for:
* [x] VS2013
* [ ] VS2012
* [ ] VS2010

### DIFF
--- a/src/gsCore/gsMemory.h
+++ b/src/gsCore/gsMemory.h
@@ -69,7 +69,7 @@ usage:
 memory::unique_ptr<int> B;
 \endcode
 */
-#if __cplusplus >= 201103 || _MSVC_LANG >= 201103
+#if __cplusplus >= 201103 || _MSC_VER >= 1600
 using std::unique_ptr;
 using std::nullptr_t;
 #else
@@ -234,7 +234,7 @@ inline std::vector<T*> release(std::vector< unique_ptr<T> >& cont)
 
 } // namespace memory
 
-#if __cplusplus >= 201103 || _MSVC_LANG >= 201103
+#if __cplusplus >= 201103 || _MSC_VER >= 1600
 /** 
     Alias for std::move, to be used instead of writing std::move for
     keeping backward c++98 compatibility
@@ -398,7 +398,7 @@ inline void copy(T begin, T end, U* result)
 
 } // namespace gismo
 
-#if __cplusplus < 201103L && _MSVC_LANG < 201103L
+#if __cplusplus < 201103L && _MSC_VER < 1600
 // Define nullptr for compatibility with newer C++
 static const gismo::memory::nullptr_t nullptr ={};
 #endif
--- a/src/gsCore/gsMemory.h
+++ b/src/gsCore/gsMemory.h
@@ -69,7 +69,7 @@ usage:
 memory::unique_ptr<int> B;
 \endcode
 */
-#if __cplusplus >= 201103 || (defined(_MSC_VER) && _MSC_VER >= 1600)
+#if __cplusplus >= 201103 || _MSVC_LANG >= 201103
 using std::unique_ptr;
 using std::nullptr_t;
 #else
@@ -234,7 +234,7 @@ inline std::vector<T*> release(std::vector< unique_ptr<T> >& cont)
 
 } // namespace memory
 
-#if __cplusplus >= 201103 || (defined(_MSC_VER) && _MSC_VER >= 1600)
+#if __cplusplus >= 201103 || _MSVC_LANG >= 201103
 /** 
     Alias for std::move, to be used instead of writing std::move for
     keeping backward c++98 compatibility
@@ -398,7 +398,7 @@ inline void copy(T begin, T end, U* result)
 
 } // namespace gismo
 
-#if (__cplusplus < 201103) && (!defined(_MSC_VER) || _MSC_VER < 1600)
+#if __cplusplus < 201103L && _MSVC_LANG < 201103L
 // Define nullptr for compatibility with newer C++
 static const gismo::memory::nullptr_t nullptr ={};
 #endif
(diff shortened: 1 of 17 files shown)


src/gsCore/gsTemplateTools.h
```diff
@@ -22,7 +22,7 @@ namespace gismo
 
 namespace util {
 
-#if __cplusplus >= 201103 || (defined(_MSC_VER) && _MSC_VER > 1700)
+#if __cplusplus >= 201103L || _MSVC_LANG >= 201103L // (defined(_MSC_VER) && _MSC_VER > 1700) // MSVC 2012
 //see also http://lists.boost.org/Archives/boost/2009/04/151209.php
 template <typename T> struct has_move_constructor
```
codewerfer: This seams to do not work with MSVC.

filiatra: I would say, then you could leave it as it was with the version check

filiatra: and leave a note that this is due to reduced support of older MSVC version

codewerfer: It doesn't work with any (newer) MSVC Version - I can only test with 2015, 2017 and 2019.

Have no idea how to find out if some class has a move constructor.
I mean, if there is any good solution, why they didn't used it for `std::is_move_constructible`?

Btw. my information now is, that `_MSVC_LANG` was [introduced with VS 2015](https://stackoverflow.com/a/46243045/11323147).
All [Versions up from VS 2010 supported parts of Cxx11](https://docs.microsoft.com/en-us/previous-versions/hh567368(v=vs.140)) - more or less. [VS 2008 should have been pure C++98](http://www.cplusplus.com/forum/windows/225309/). There seams to exist no way to force back to C++98 or 11. The `/std:xxx` command was [introduces in VS 2017](https://docs.microsoft.com/de-de/cpp/build/reference/std-specify-language-standard-version?view=vs-2015) and supports only `c++14`, `c++17`, `c++latest` and `c++20` when its implementation is completed - changing `CXX_STANDARD` in CMake does exactly this.

[`Rvalue References`](https://docs.microsoft.com/en-us/previous-versions/hh567368(v=vs.140)#rvalue-references) are supported in some different implementation versions up from VS 2010 [(v2.0)](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2009/n2844.html), VS 2012 and 2013 [(v2.1)](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html) and VS 2015 [(v3.0)](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3053.html). Not sure what's the first one we support - please look into the article(s).

---

I will move back some of the `_MSC_VER` code for older compilers than VS 2015 (`_MSC_VER 1900`). But newer features (14, 17) should be enabled with `_MSVC_LANG` or setting `__cplusplus` to the correct value with `add_compile_options("/Zc:__cpluscplus")`.

I don't like that I have no way here to test it with older compilers, can you (@filiatra) set up a AppVeyor build for 2010, 2012, 2013?

src/gsUtils/gsUtils.h
```diff
@@ -268,7 +268,7 @@ size_t hash_range(T const * start, const T * const end)
     return seed;
 }
 
-#if __cplusplus >= 201703L
+#if __cplusplus >= 201703L || _MSVC_LANG >= 201703L
```
codewerfer: correct

src/gsCore/gsTemplateTools.h
filiatra: OK then leave the stuff as they are. For newer features we should use this check:

#if (defined(_MSVC_LANG) && _MSVC_LANG > 201402) || __cplusplus >= 201402

filiatra: I als don't know how to chec has_move_constructor for MSVC. YOu can disable the test for this compiler and leave a comment inside the test thas explains that the function does not work correctly in MSVC.

codewerfer: > OK then leave the stuff as they are. For newer features we should use this check:
> 
> #if (defined(_MSVC_LANG) && _MSVC_LANG > 201402) || __cplusplus >= 201402

It's enough to check for `#if _MSVC_LANG >= 201402L || __cplusplus >= 201402` cause all ["remaining identifiers and keywords, ... are replaced with the pp-number 0"](https://github.com/cplusplus/draft/blob/c635711cdd81346ad41c7861adb8035176fa236f/source/preprocessor.tex#L357-L375), therefore only for `_MSVC_LANG < 201703L` we need a `defined(_MSVC_LANG)`.

src/gsCore/gsMemory.h
```diff
@@ -398,7 +398,7 @@ inline void copy(T begin, T end, U* result)
 
 } // namespace gismo
 
-#if (__cplusplus < 201103) && (!defined(_MSC_VER) || _MSC_VER < 1600)
+#if __cplusplus < 201103L && _MSC_VER < 1600
```
codewerfer: | Compiler | __cplusplus | < 201103L | _MSC_VER | < 1600 | && |
| --- | --- | --- | --- | --- | --- |
| GCC 97 | 199711L | true | 0 | true | true |
| GCC 11+ | 201103L | false | 0 | true | false |
| MSVC 2008 | 199711L | true | 1500  | true | true |
| MSVC 2010+ | 199711L | true | 1600 | false | false |

src/gsCore/gsTemplateTools.h
codewerfer: > OK then leave the stuff as they are.

[Current `stable` doesn't compile with `VS 2013` or lower.](https://cdash.ricam.oeaw.ac.at/buildSummary.php?buildid=348207)
So, we have to fix it anyway - or kick out "support" for older compilers.

filiatra:
warns: https://cdash.ricam.oeaw.ac.at/viewBuildError.php?type=1&buildid=350477

filiatra:
OK I will merge this and I let the warnings of the other PR

PR #332, opened 2019-07-21
Gs comp flow extension

mmoelle1:
Several updates:

1.   Removed `gsFdbb` extension from stable branch and made it a submodule of the `gsCompFlowExtension` where it is used; 
2.  Added optional parameter to `gismo_add_executable` macro to specify name of the executable explicitly; 
3.  Added additional routines to `gsMpi`.


src/gsMpi/gsMpi.h
```diff
@@ -175,10 +176,37 @@ class gsMpi
 #   ifdef GISMO_WITH_MPI
         if( !initialized() )
         {
+#ifdef _OPENMP
```
filiatra: It would be preferable to get the MPI_INIT_THREAD as follows:

Thread support is by default 'MPI_THREAD_SINGLE'.  This may be overridden with the
GISMO_MPI_THREAD_LEVEL environment variable.  If set, this variable
replaces the default 'MPI_THREAD_SINGLE' value.  The following values
are allowed

0: Corresponds to 'MPI_THREAD_SINGLE'

1: Corresponds to 'MPI_THREAD_FUNNELED'

2: Corresponds to 'MPI_THREAD_SERIALIZED'

3: Corresponds to 'MPI_THREAD_MULTIPLE'

Issue #348, opened 2019-11-05
Parametric mesh is plotted incorrectly for a scalar-valued THB.


dominik-mokris-mtu:
Let's define a scalar-valued THB-spline and plot it in Paraview.

```cpp
#include <gismo.h>

using namespace gismo;

int main()
{
    gsKnotVector<real_t> knots(0,1,1,2);
    gsTensorBSplineBasis<2> tbasis(knots, knots);
    gsMatrix<real_t> coefs(9, 1);
    coefs << 0, 0, 0.1, 0, 0.2, 0.3, 0.2, 0.2, 0.1;
    gsTHBSpline<2> thb(tbasis, coefs);

    std::vector<unsigned> boxes(5);
    boxes[0] = 1;
    boxes[1] = 0;
    boxes[2] = 0;
    boxes[3] = 1;
    boxes[4] = 1;
    thb.basis().refineElements_withCoefs(thb.coefs(), boxes);

    gsWriteParaview(thb, "thb", 1000, true, false);
    return 0;
}
```
The element mesh is badly formed,
![gismo_physical_old](https://user-images.githubusercontent.com/43567793/68216941-a99aee80-ffd9-11e9-8d84-1936234ac5ab.png)
since the expected result would be something like this:
![gismo_physical_new](https://user-images.githubusercontent.com/43567793/68216994-bf101880-ffd9-11e9-9b4d-ad972d848e66.png)

I have a bugfix, which is a part of the pull request #349.

dominik-mokris-mtu:
Also solved in #349 , thanks to everyone who helped!

Issue #347, opened 2019-10-30
gsWriteParaview: parametric mesh too detailed, thus taking a lot of time


dominik-mokris-mtu:
We've had a discussion with @Sandra-Merchel-MTU recently about why `gsWriteParaview` was taking so much time on her computer. We have discovered the following problem.

When you call
```cpp
    gsWriteParaview(myTHB, "filename", 1000, true);
```
(because you want to see the parametric mesh), it will call `gsWriteSingleCompMesh` with `npts` equal to 1000, which is much more than the default 8. Consequently, you evaluate each edge of the parametric mesh in 1000 points, taking a lot of time (the THB face took from the MWE below took 20m, 59s on my computer). This does not seem necessary, as it results in the parametric mesh much finer than the rendering of the underlying geometry.

**MWE**
```cpp
#include <gismo.h>

using namespace gismo;

int main()
{
    gsFileData<real_t> data("surfaces/thbs_face_3levels.xml");
    gsTHBSpline<2>::uPtr THB = data.getFirst<gsTHBSpline<2> >();

    gsStopwatch time;
    time.restart();
    gsWriteParaview(*THB, "thb_out", 1000, true);
    time.stop();
    gsInfo << "gsWriteParaview took " << time << std::endl;
    return 0;
}
```
**Possible solutions**

1. We change the call of `gsWriteSingleCompMesh` to use `8` instead of `npts`.
2. We change the call of `gsWriteSingleCompMesh` to use some reasonable value derived from `npts`. Taking something of the order of `math::sqrt(mesh.numberOfElements() / npts)` should lead to a similar resolution of the parametric mesh and the underlying geometry.
3. We proclaim it a problem of the user. (-;

dominik-mokris-mtu:
Good. I found one more thing I want to improve in `gsWriteParaview`. Once it's ready for a pull request, I will add this little enhancement there.

dominik-mokris-mtu:
Solved in #349 , thanks to everyone who helped!

PR #349, opened 2019-11-05
Paraview mesh bug

dominik-mokris-mtu:
This pull request does three things:
1. It implements the enhancement discussed in issue #347 (commit 3b13568d17dec7800b61d7dd1314f8fe73ba9ba8).
2. It contains a bugfix to issue #348 (commit 7f4d38546d3878bbb45ba42a73caa5ee12fbfcb4).
3. Documentation of several functions in `gsMesh`, which were used for the first two points, has been made and updated (commit acaaf6fb7ffb8da8fc1738b92437de0ad81e8762).

# The famous checklist:
  **Yes**, in the form of issues #347 and #348 .
  There is not much new code but existing Doxygen has been improved.
  **No**. Both of the issues improve the visualisation with Paraview, which would be difficult to capture in an automated test.
-----

### DIFF
--- a/src/gsCore/gsGeometry.hpp
+++ b/src/gsCore/gsGeometry.hpp
@@ -129,7 +129,7 @@ void gsGeometry<T>::evaluateMesh(gsMesh<T>& mesh) const
 
     // For all vertices of the mesh, push forward the value by the
     // geometry mapping
-    if (1==gDim && 3<pDim) // Plot a graph
+    if (1==gDim && 3>pDim) // Plot a graph
         for (size_t i = 0; i!= mesh.numVertices(); ++i)
         {
             eval_into( mesh.vertex(i).topRows(pDim), tmp );
--- a/src/gsCore/gsGeometry.hpp
+++ b/src/gsCore/gsGeometry.hpp
@@ -132,7 +132,7 @@ void gsGeometry<T>::evaluateMesh(gsMesh<T>& mesh) const
     for (size_t i = 0; i!= mesh.numVertices(); ++i)
     {
         eval_into( mesh.vertex(i).topRows(pDim), tmp );
-        mesh.vertex(i).topRows( gDim ) = tmp;
+        mesh.vertex(i).bottomRows( gDim ) = tmp;
     }
 }
 template<class T>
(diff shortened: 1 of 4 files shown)


src/gsCore/gsGeometry.hpp
```diff
@@ -129,11 +129,19 @@ void gsGeometry<T>::evaluateMesh(gsMesh<T>& mesh) const
...
+            eval_into( mesh.vertex(i).topRows(pDim), tmp );
+            mesh.vertex(i).middleRows(pDim, gDim) = tmp;
+        }
+    else // Plot mesh on a mapping
+        for (size_t i = 0; i!= mesh.numVertices(); ++i)
+        {
+            eval_into( mesh.vertex(i).topRows(pDim), tmp );
+            mesh.vertex(i).topRows(gDim) = tmp;
+        }
+
```
dominik-mokris-mtu: What was wrong with using `bottomRows`? I tested it for
f: [0,1]^2 -> R^1
and
f: [0,1]^2 -> R^3
and it worked in both cases.

The `middleRows` solution gives a wrong result in the test case from issue #348, at least on my computer.

dominik-mokris-mtu:
I'm sorry but the commit 6f09d358379c3827466e3113871d0a01316ced5b restored the buggy behaviour from issue #348, see my comment in the commit.

src/gsCore/gsGeometry.hpp
filiatra: I think that bottomRows would not be correct for case
 f: [0,1]^1 -> R^1
Essentially, middleRows(pDim, gDim) should be the same result as bottomRows(gDim) for the case #348, I will check again..

dominik-mokris-mtu: I was also surprised. My guess was that `middleRows(pDim+1, gDim)` would be the same as the `bottomRows` but even that did not behave correctly. It is of course possible that I have made a mistake somewhere on the way but I have double-checked and found nothing wrong.

dominik-mokris-mtu: One more thought occurred to me: what is actually the expected outcome for
f: [0,1]^1 -> R^1?

What is the element mesh supposed to depict? Does it make sense to do something like
`gsWriteParaview(fn, string, 100, true)` at all?

filiatra:
I think that this works now, I let you merge if your are fine with this.
I need the if/else condition because the function should also work in different scenarios..

PR #355, opened 2020-02-14
Bugfix: a forgotten swap of parameters from issue #290

dominik-mokris-mtu:
# Short explanation
When I was fixing the swapped _u_- and _v_-directions in Parasolid export (issue #290), I missed one of them. The current PR fixes the problem.

# Longer explanation
Parasolid and Gismo store coefficients in a different order. In the past, this meant that the Parasolid export was returning a geometrically correct surface but with _u_ and _v_ swapped. In issue #290, I fixed this for all our calls of the functions in `gsWriteParasolid.hpp`. However, I did not realize that there are functions in the file that call other functions from the _same_ file and did not fix the transposition for those. The current PR fixes this omission.

# Smallprint
approval) from @gismo/admins.

and free of bugs.

request:
  No, it is a bugfix.
  No. We might consider adding some unittests in the future, though.
-----

### DIFF
--- a/extensions/gsParasolid/gsWriteParasolid.hpp
+++ b/extensions/gsParasolid/gsWriteParasolid.hpp
@@ -850,7 +850,7 @@ bool exportTHBsurface( const gsTHBSpline<2, T>& surface,
         //std::cout << "bspline coef size: " << bspline.coefs().rows() << std::endl;
         makeValidGeometry(surface, bspline);
 
-        //gsWriteParaview(bspline,"paraviewtest" + std::to_string(box),1000,true,true);
+        //gsWriteParaview(bspline,"paraviewtest" + util::to_string(box),1000,true,true);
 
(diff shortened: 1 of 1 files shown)


extensions/gsParasolid/gsWriteParasolid.hpp
```diff
@@ -850,7 +850,7 @@ bool exportTHBsurface( const gsTHBSpline<2, T>& surface,
         //std::cout << "bspline coef size: " << bspline.coefs().rows() << std::endl;
         makeValidGeometry(surface, bspline);
 
-        //gsWriteParaview(bspline,"paraviewtest",1000,true,true);
+        //gsWriteParaview(bspline,"paraviewtest" + std::to_string(box),1000,true,true);
```
codewerfer: `util::to_string` for C++98 compatibility

filiatra:
OK

PR #367, opened 2020-06-17
Update CMakeLists.txt

mmoelle1:
When Trilinos is compiled with SuperLU and/or UMFPACK enabled the corresponding header files need to be available in the include directories. Therefore, the Trilinos package is included after all others with an extra refresh of the include directories before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,13 +276,6 @@ if(GISMO_WITH_UNUM)
   #include_directories(${UNUM_INCLUDE_DIR})
 endif(GISMO_WITH_UNUM)
 
-if(GISMO_WITH_TRILINOS)
-  add_subdirectory(extensions/gsTrilinos)
-  set (GISMO_INCLUDE_DIRS ${GISMO_INCLUDE_DIRS} ${TRILINOS_INCLUDE_DIR}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -300,6 +293,16 @@ endif()
...
+if(GISMO_WITH_TRILINOS)
+  add_subdirectory(extensions/gsTrilinos)
+  set (GISMO_INCLUDE_DIRS ${GISMO_INCLUDE_DIRS} ${TRILINOS_INCLUDE_DIR}
+  CACHE INTERNAL "${PROJECT_NAME} include directories")
+  #include_directories(${TRILINOS_INCLUDE_DIR})
+endif(GISMO_WITH_TRILINOS)
+
+#third time
+include_directories(${GISMO_INCLUDE_DIRS})
+
```
filiatra, line 305: The "second time" could be removed safely probably?

filiatra (approved):

CMakeLists.txt
mmoelle1: No, the "second time" is needed so that Trilinos can find "SuperLU" and "UMFPACK".

PR #368, opened 2020-06-17
Update CMakeLists.txt

mmoelle1:
Updated Trilinos to version 12-14-1.

### DIFF
--- a/extensions/gsTrilinos/CMakeLists.txt
+++ b/extensions/gsTrilinos/CMakeLists.txt
@@ -35,48 +35,40 @@ aux_tmpl_header_directory(${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_NAME}_HPPFILES)
 # Apply same configuration as G+Smo
 include(gsConfig)
 
-# Look for MPI (now: handled by GISMO_WITH_MPI)
-#find_package(MPI REQUIRED)
-#include_directories(SYSTEM ${MPI_INCLUDE_PATH})
-
-# Look for trilinos library
-#find_package(TRILINOS REQUIRED) # for testing user-provided mode
+# Look for pre-installed Trilinos libraries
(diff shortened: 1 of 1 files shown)


extensions/gsTrilinos/CMakeLists.txt
```diff
@@ -35,48 +35,40 @@ aux_tmpl_header_directory(${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_NAME}_HPPFILES)
...
           URL_MD5 ${TRILINOS_MD5}
           URL https://github.com/trilinos/Trilinos/archive/trilinos-release-${TRILINOS_VER}.tar.gz
           CMAKE_ARGS   -DCMAKE_BUILD_TYPE=Release
                        -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
                        -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                       #-DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
                        -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
                        -DBUILD_SHARED_LIBS:BOOL=ON
-                       -DTrilinos_ENABLE_CXX11=${Trilinos_ENABLE_CXX11}
+                       -DTrilinos_ENABLE_CXX11=ON
```
filiatra, line 66: the flag "Trilinos_ENABLE_CXX11" could be on by default, it did not work as expected?

mmoelle1: Basically all features from Trilinos that we need in G+Smo require C++11, at least in version 12-14-1. If we allow disabling C++11 support then we need to resort to an older Trilinos version.

filiatra (approved):
OK

PR #377, opened 2020-10-31
Fixed bug in CircleCI pipeline

mmoelle1:
CircleCI pipeline fails because brew tries to install `git` which is already installed.


filiatra:
now working on gitlab_ci

PR #396, opened 2020-12-10
added neumann conditions

hverhelst:
small change in kirchhoff example. Neumann assembly was lost.

### DIFF
--- a/doc/kirchhoff-love_example.dox
+++ b/doc/kirchhoff-love_example.dox
@@ -9,7 +9,7 @@ In this example, the Kirchhoff-Love shell is implemented. References for this im
 
 The variational formulation for the Kirchhoff-Love shell is:
 \f{eqnarray*}{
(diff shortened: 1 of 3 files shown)


filiatra (approved):

hverhelst:
The previous commit was not (yet) the full fix. Now I added the otangent expression since that was needed (have verified it somewhere else). The `tangent_expr` does exist in `gsExpressions.h`, but that only works for 2D. Would it be nice to combine the 2D and 3D tangent expressions (so `tangent_expr` and `otangent_expr)` to one single expression?

filiatra:
> Would it be nice to combine the 2D and 3D tangent expressions (so `tangent_expr` and `otangent_expr)` to one single expression?

Indeed it would be nice to combine both to a single expression. It is easy to just add an if/else inside the eval(.) of the existing tangent_expr.

filiatra (changes requested):
For the cross product:
gsMatrix<> a(3,1), b(3,1);

a.cross(b) does not work but something like

a.head<3>().cross(b.head<3>())  should do the trick.

hverhelst (commented):
Merged the tangent and otangent expressions now. The return type of the eval changed from MatExprType to gsMatrix<Scalar>

filiatra (approved):

PR #401, opened 2020-12-11
Set CMake 2.8.12 as minimum version

mmoelle1:
CMake 3.19 issues a deprecation warning for `cmake_minimum_required` smaller than 2.8.12.


filiatra:
I would like to see if we can silence that warning instead of changing this.. 
There are some builds here on older systems: https://build.opensuse.org/package/show/home:filiatra/gismo
and it is a pity to drop support due to this warning..

mmoelle1:
I think that disabling the warning is not an option. 

This is what I see with CMake 3.19.
```
CMake Deprecation Warning at /Users/mmoller/codes/gismo.trilinos/external/Trilinos/cmake/tribits/core/config_tests/fmangle/CMakeLists.txt:40 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

CMake 2.8.12 is from 2013. Ubuntu 16.04 ships with CMake 3.5.

filiatra:
I also see that we do not have any alternatives. But we should grep this all over the codebase, there are many more of these statements in sub-projects

filiatra (approved):
OK I see that it is done all the way

PR #402, opened 2020-12-14
Improve support for unclamped knots (fix #10 and fix #350)

### LINKED ISSUE #10: hierarchical b-spline refinement doesn't work with uniform knotvectors?
Hello everybody,

I would like to use the hierarchical b-spline structure of Gismo. But i got the following issue:
I defined a 2D Domain using uniform knotvectors with multiplicity = 1 at the two end knots:

`  real_t a = -0.5;                   // starting knot`
`real_t b = 1.5;                   // ending knot`
   ` unsigned interior = 7;     // number of interior knots`
    `int      degree = 2;`
    `unsigned mult = 1;  // multiplicity at the two end knots`
   ` gsKnotVector<> kv(a,b,interior, mult);`
    `kv.set_degree(degree);`

Then i got the following knotvector:
`[ -0.5 -0.25 0 0.25 0.5 0.75 1 1.25 1.5 ]`

Now i create the hierarchical b-spline structure :
`gsTensorBSplineBasis<2,real_t> tens( kv, kv);`
`gsHBSplineBasis<2,real_t> hb(tens);`

and try to refine the square [0, 0]×[0.5, 0.5]
    `std::vector<unsigned> box;`
    `box.push_back( 1 );`
   ` box.push_back( 0 );`
    `box.push_back( 0 );`
    `box.push_back( 4 );`
    `box.push_back( 4 );`
   `hb.refineElements(box);`

After exporting to a paraview file. 
`gsWriteParaview(hb, "test_hb_refined" );`

The region which was refined ist [0, 0] x [0.25, 0.25] and not [0, 0]×[0.5, 0.5]. Did i use the wrong indices for the box of refined region or the "refineElements" function only works with knotvectors with multiplicity = degree + 1 at the endknots?

Many thanks in advance!
skyforcen73

### LINKED ISSUE #350: Function refineElements does not work for gsTHBSplineBasis without clamped knots.
The following code should construct a THB-spline basis without clamped knots and refine it.

```cpp
#include <gismo.h>

using namespace gismo;

int main()
{
    // Knot vector for a quadratic basis without clamped knots.
    gsKnotVector<real_t> kv(-0.5,1.5,7,1);
    kv.set_degree(2);

    gsTensorBSplineBasis<2, real_t> ten(kv, kv);
    gsTHBSplineBasis<2> thb(ten);

    std::vector<index_t> box;
    box.push_back( 1 );
    box.push_back( 0 );
    box.push_back( 0 );
    box.push_back( 8 );
    box.push_back( 8 );
    thb.refineElements(box);

    gsInfo << thb << std::endl;

    return 0;
}
```
However, it does not go through:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  GISMO_ASSERT `i>=0 && i<m_size` 
/home/cpfscae07/disk00100/disk00101/ya10813/gismo/external/Eigen/src/SparseCore/SparseVector.h, line 127 (coeffRef)
Aborted
```

After quite some chase throught the source code I was able to pinpoint the cause of the behaviour to the lines 783 and 784 in gsKnotVector.hpp.
```cpp
    if (0!=l) trimLeft (l);
    if (0!=r) trimRight(r);
```
These two cause the level 1 knot vector of to be
```
[ -0.25 -0.1875 -0.125 -0.0625 0 ... 1 1.0625 1.125 1.1875 1.25 ] (deg=2, size=25, minSpan=0.0625, maxSpan=0.0625)
```
instead of
```
[ -0.5 -0.4375 -0.375 -0.3125 -0.25 ... 1.25 1.3125 1.375 1.4375 1.5 ] (deg=2, size=33, minSpan=0.0625, maxSpan=0.0625)
```
that I would expect. That is, the _domain_ of the knot vector (the interval between the (`deg+1`)-st from the beginning and `(deg+1)`-st knot from the end) remains the same but the _first_ and _last_ knots shrink closer to each other.

Uncommenting these two lines not only alleviated the problem but it also solved issue #10 (which is how I got to this problem in the first place).

What shall we do? My suggestion would be to get rid of those two lines but I suppose they are there for a reason.

filiatra:
The idea is to number the unique left ghost knots in a knot vector with negative integers.


osmihran:
Hi! Thanks, I will have a look at it in the coming days.

src/gsHSplines/gsHDomainBoundaryIterator.h
```diff
@@ -208,22 +208,22 @@ class gsHDomainBoundaryIterator: public gsDomainIterator<T>
             {
                 if ( par )
                 {
-                    m_breaks[dim].push_back(kv.uValue(end-1));
-                    m_breaks[dim].push_back(kv.uValue(end  ));
+                    m_breaks[dim].push_back( kv(end-1) );
+                    m_breaks[dim].push_back( kv(end  ) );
 
                     //  = knotValues.begin() + end -1;
```
osmihran, line 214: Can we remove these comments? Also, there is another one a few lines up.

src/gsHSplines/gsHDomainBoundaryIterator.h
```diff
@@ -208,22 +208,22 @@ class gsHDomainBoundaryIterator: public gsDomainIterator<T>
             {
```
osmihran, line 208: I haven't found any place, where updateLeaf() is called. How do we know it is correct?

src/gsHSplines/gsTHBSplineBasis.hpp
```diff
@@ -53,24 +53,25 @@ void gsTHBSplineBasis<d,T>::representBasis()
...
     {
-        unsigned level = this->levelOf(j);
+        index_t level = this->levelOf(j);
         index_t tensor_index = this->flatTensorIndexOf(j, level);
 
         // element indices
-        gsMatrix<index_t, d, 2> element_ind(d, 2);
         this->m_bases[level]->elementSupport_into(tensor_index, element_ind);
 
         // I tried with block, I can not trick the compiler to use references
```
osmihran, line 66: Is this comment still relevant?

src/gsHSplines/gsHDomainBoundaryIterator.h
osmihran, line 206: I haven't found any place where updateLeaf() is called. How do we know if it's correct or not?

src/gsHSplines/gsHDomainBoundaryIterator.h
osmihran: Can we remove these comments? And maybe the one on line 204 as well?

src/gsHSplines/gsHDomainBoundaryIterator.h
```diff
@@ -208,22 +208,22 @@ class gsHDomainBoundaryIterator: public gsDomainIterator<T>
...
+                    m_breaks[dim].push_back( kv(end  ) );
 
                     //  = knotValues.begin() + end -1;
                     //  = knotValues.begin() + end   ;
                 }
                 else
                 {
-                    m_breaks[dim].push_back(kv.uValue(start));
-                    m_breaks[dim].push_back(kv.uValue(start+1));
+                    m_breaks[dim].push_back( kv(start)   );
```
osmihran, line 214: It seems that you have replaced "kv.uValue" with "kv". I don't see yet, why that's the correct thing to do.

src/gsHSplines/gsTHBSplineBasis.hpp
```diff
@@ -1470,11 +1469,11 @@ void gsTHBSplineBasis<d, T>::findNewAABB(const std::vector< std::vector<T> >& po
     aabb.resize(4);
     for (unsigned i = 0; i != kv0.uSize(); i++)
     {
-        if (kv0.uValue(i) <= minX)
```
osmihran, line 1473: Again, I don't see yet, why this is the right thing to do.

src/gsHSplines/gsHDomainIterator.h
```diff
@@ -164,7 +164,7 @@ class gsHDomainIterator: public gsDomainIterator<T>
 
             m_breaks[dim].clear();
             for (unsigned index = start; index <= end; ++index)
-                m_breaks[dim].push_back(kv.uValue(index));
```
osmihran, line 167: Again, I don't see yet, why this is the right thing to do.

src/gsNurbs/gsKnotIterator.h
```diff
@@ -77,15 +78,14 @@ class gsUKnotIterator
        Optionally the iteration starts from from the knot with unique
        index (i.e. without repetitions) equal to \a upos
      */
-    explicit gsUKnotIterator(knotVector & KV, const mult_t upos = 0)
-    : m_mlt ( KV.multSumData() ),
-      m_raw ( KV.data()        ),
-      m_upos( upos             )
+    explicit gsUKnotIterator(knotVector & KV, const mult_t upos = 0, const index_t s = 0)
```
osmihran, line 84: The meaning of "s" should be mentioned in the doxygen documentation.

src/gsNurbs/gsKnotIterator.h
```diff
@@ -259,7 +259,14 @@ class gsUKnotIterator
        \brief Returns the index counted without repetitions (i.e. the
        unique index) of the current knot
      */
-    mult_t uIndex() const {return m_upos;}
+    mult_t uIndex() const {return m_upos-m_sh;}
+
+    /**
+       \brief Returns the number of knots counted without repetitions
+       thet are on the left of the current knot (differs from uIndex
```
osmihran: "thet" -> "that"
And maybe a full stop at the end?

src/gsNurbs/gsKnotIterator.h
```diff
@@ -360,13 +367,13 @@ class gsKnotIterator
        Optionally the iteration starts from from the first appearance
        of the knot with unique index (i.e. without repetitions) equal to \a upos
     */
```
osmihran, line 375: Again, "s" should be explained in the doxygen comment.

src/gsNurbs/gsKnotIterator.h
```diff
@@ -456,14 +463,26 @@ class gsKnotIterator
...
         GISMO_ENSURE(m_pos+a>=0 && m_pos+a+1 < m_dbg, "Invalid access to non-existent knot.");
 #       endif
         return m_uit.m_raw[m_pos+a];
     }
 
+    /**
+       \brief Returns the value of the knot which is \a a positions
+       after the current knot (counted without repetitions).
+    */
+    reference operator() (ptrdiff_t a) const
```
osmihran, line 484: Nice!

src/gsNurbs/gsKnotIterator.h
```diff
@@ -586,7 +605,14 @@ class gsKnotIterator
        \brief Returns the index counted without repetitions (i.e. the
        unique index) of the current knot
      */
-    mult_t uIndex() const {return m_uit.m_upos;}
+    mult_t uIndex() const {return m_uit.uIndex();}
```
osmihran, line 614: Why?

src/gsNurbs/gsKnotIterator.h
```diff
@@ -586,7 +605,14 @@ class gsKnotIterator
...
+
+    /**
+       \brief Returns the number of knots counted without repetitions
+       thet are on the left of the current knot (differs from uIndex
```
osmihran: "thet" -> "that"

src/gsNurbs/gsKnotVector.h
```diff
@@ -239,13 +239,23 @@ class gsKnotVector : public gsDomain<T>
...
     }
 
+    /// Provides the knot with unique index \a i
+    const T& operator()(const mult_t i) const
+    {
+        return *( this->ubegin()+(numLeftGhosts()+i) );
+    }
+
+    /// Provides the knot with unique index \a i
+    inline T  uValue(const size_t & i) const
```
osmihran, line 256: Is this a simpler implementation of "uValue" or does it return different values than the old one?

src/gsNurbs/gsKnotVector.h
```diff
@@ -370,6 +380,22 @@ class gsKnotVector : public gsDomain<T>
     /// Removes the right-most \a numKnots from the knot-vector
     void trimRight(const mult_t numKnots);
 
+    /// Computes the number of left ghosts, i.e., of the knots to the
+    /// left of the domain beginnning.
```
osmihran: Too many "n" (-;

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -302,7 +304,7 @@ void gsKnotVector<T>::insert( T knot, mult_t mult )
     const mult_t fa = uit.firstAppearance();
 
     // update multiplicity sums
-    nonConstMultIterator upos = m_multSum.begin() + uit.uIndex();
+    nonConstMultIterator upos = m_multSum.begin() + uit.uCardinalIndex();
```
osmihran, line 307: Why?

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -632,7 +634,7 @@ void gsKnotVector<T>::initUniform( T first,
 
     const T h = (last-first) / (interior+1);
 
-    for(unsigned i = m_deg - mult_ends + 1, j=0; i!= 0; --i, ++j)
```
osmihran, line 635: Was this a bug? If yes, shouldn't we make a unit test to prevent it from returning?

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -632,7 +634,7 @@ void gsKnotVector<T>::initUniform( T first,
...
+    for(unsigned i = m_deg - mult_ends + 1, j=1; i!= 0; --i, ++j)
```
osmihran, line 637: Why "unsigned" and not "mult_t" or "index_t"?

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -695,7 +697,7 @@ typename gsKnotVector<T>::uiterator
 gsKnotVector<T>::uFind( const T u ) const
 {
     GISMO_ASSERT(size()>1,"Not enough knots."); // todo: check() --> size() > 2*m_deg+1
-    GISMO_ASSERT(inDomain(u), "Point outside active area of the knot vector");
+    GISMO_ASSERT(inDomain(u), "Point "<<u<<" outside active area of the knot vector");
```
osmihran: Nice!
I would also vote for spaces between "<<" and "u".

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -991,13 +993,11 @@ template< typename T>
 void gsKnotVector<T>::supportIndex_into(const mult_t& i,
```
osmihran, line 993: Is this "only" a smarter implementation or does it change the results?

If the latter, we should change the documentation. Also, it is used in gsSmoothPatches in unsupported. Aren't we breaking it?

src/gsNurbs/gsTensorBSplineBasis.h
```diff
@@ -371,8 +371,10 @@ class gsTensorBSplineBasis : public gsTensorBasis<d,T>
 
         for (short_t dim = 0; dim < d; ++dim)
         {
-            Self_t::component(dim).knots().supportIndex_into(ti[dim], tmp_vec);
-            result.row(dim) = tmp_vec.row(0);
+            const gsKnotVector<T> & kv = Self_t::component(dim).knots();
+            kv.supportIndex_into(ti[dim], tmp_vec);
+            result.row(dim).noalias() =
+                tmp_vec.cwiseMax(0).cwiseMin(kv.numElements());
```
osmihran, line 377: The reason for the cwise trick is that supportIndex_into can now (due to the handling of ghosts) sometimes return things below zero?

unittests/gsKnotVectors_test.cpp
```diff
@@ -260,12 +260,24 @@ SUITE(gsKnotVectors_test_2)
...
-            CHECK( KV.uFind(0.50).uIndex() == 4 );
-            CHECK( KV.uFind(0.55).uIndex() == 4 );
-            CHECK( KV.uFind(0.99).uIndex() == 5 );
-            CHECK( KV.uFind(1.00).uIndex() == 5 );
+            CHECK( KV.numLeftGhosts() == 1 );
+            CHECK( KV[0] == KV(-1) ); // unique indices of left ghosts are negative
+            CHECK( KV[1] == KV(0)  );
+            CHECK( KV[10] == KV(7) );
+
+            CHECK( KV.uFind(0.00).uIndex() == 0 );
```
osmihran, line 288: So what was called "uIndex" is now called "uCardinalIndex"? How do we know we haven't broken any of the approx. 85 invocations?
Honestly, I am not very happy about such renaming of functions. Is the new naming more logical?

osmihran (commented):
First, let me thank you for pushing it, you did a lot of good work.

I have left a few pedantic comments, can we talk about them before merging?

Also, where do all the spaces at the line ends come from? (=

src/gsHSplines/gsHDomainBoundaryIterator.h
filiatra: yes these comments can be removed

src/gsHSplines/gsHDomainBoundaryIterator.h
filiatra: I added a new operator  kv(index) that does the same thing as uValue.
There is no difference between the two, so I start using here the operator instead.

src/gsHSplines/gsHDomainIterator.h
filiatra: kv(index) returns the (signed) unique index, same as uValue.
Note that ghost knots have unique index < 0.

src/gsHSplines/gsTHBSplineBasis.hpp
filiatra: I think it is, we could change this to a column reference to avoid the extra copy, but it is not a big thing

src/gsHSplines/gsTHBSplineBasis.hpp
filiatra: Just renaming the function.

src/gsNurbs/gsKnotIterator.h
filiatra: The unique index is provided by the gsUKnotIterator.
Here we call the dedicated function instead of accessing the internals of gsUKnotIterator.
In practice: the gsUKnotIterator applies a possible shift for the case of ghost knots.

src/gsNurbs/gsKnotVector.h
filiatra: The uValue and the () operator are the same thing.
The implementation is updated to account for ghost knots. If there are no ghost knots it returns the same values as before.
If there are ghost knots, the implementation is changed: ghosts are assigned negative indices, so for negative i we get the corresponding ghost knot.

src/gsNurbs/gsKnotVector.hpp
filiatra: The cardinal index is the index of the knot in the vector without multiplicitities. If we have ghost knots, the unique index differs from the cardinal index (the cardinal index always starts from 0 for the leftmost knot).
Here we need a cardinal index and not the unique index.
Note that cardinal=unique index if there are not ghost knots.

src/gsNurbs/gsKnotVector.hpp
filiatra: Yes this was a bug (for non-open knot vectors). We could add more tests for non-open knot vectors, you are welcome to start!

src/gsNurbs/gsKnotVector.hpp
filiatra: It matches the type of the arguments (eg. interior). Could be changed but one should also change the arguments as well. Works fine for now..

src/gsNurbs/gsKnotVector.hpp
filiatra: For open knot vectors no change.
For vectors with ghost knots, they get negative indices, so we get different results.
The doc is OK, actually to keep the function doing the same thing I extended it to non-open knot vectors (which have knots with negative unique indices).

gsSmoothPatches in unsupported is assuming and using open knot vectors, so no change in this respect.

src/gsNurbs/gsTensorBSplineBasis.h
filiatra: Yes, exactly.

unittests/gsKnotVectors_test.cpp
filiatra: No uIndex is not called uCardinalIndex.
In case of open knot vectors uIndex coincides with uCardinalIndex.
In this test (and in most invocations) there was an assumption that the cardinalindex and the unique index coincide always. Now the implementation is extended so I added separate tests for uIndex and uCardinalIndex. The new naming (but no renaming here!) is inteded to be logical... Is it?

filiatra:
> First, let me thank you for pushing it, you did a lot of good work.

thnx have a look at the updates

> I have left a few pedantic comments, can we talk about them before merging?

See my replies inline!

> Also, where do all the spaces at the line ends come from? (=

I do not see them.. which file/line you mean?

src/gsHSplines/gsHDomainBoundaryIterator.h
filiatra: It is used in many places inside the class.

osmihran: My bad, I must have been looking at a wrong place.

src/gsNurbs/gsKnotVector.hpp
osmihran:
```suggestion
    GISMO_ASSERT(inDomain(u), "Point "<< u <<" outside active area of the knot vector");
```

src/gsHSplines/gsHDomainBoundaryIterator.h
osmihran: Done in commit feb7657710908cd365f20c56a7c92dc758921f8d

src/gsNurbs/gsKnotVector.hpp
osmihran: I added at least one in e007f515a28da992779cb8a8625cf7395644faaf. Please, review!

osmihran (approved):
If you are happy with the new unittest that I have added (and all CI tests pass), we are done, I think!

osmihran:
> > Also, where do all the spaces at the line ends come from? (=
> 
> I do not see them.. which file/line you mean?

There was quite a lot of empty spaces on the ends of lines that you have (rightly) removed. No clue, how they got there; they might be artifacts of encoding or whatever. (-;

filiatra:
> If you are happy with the new unittest that I have added (and all CI tests pass), we are done, I think!

Looks good!
OK then I will merge it hopefully this year :)

osmihran:
> > If you are happy with the new unittest that I have added (and all CI tests pass), we are done, I think!
> 
> Looks good!
> OK then I will merge it hopefully this year :)

I knew it was worth reviewing today! ^^

PR #409, opened 2021-01-08
Fixed gsMSplines

hverhelst:
gsSmoothPatches/gsCompositeBasis.h*      ->gsMSplines/gsMappedBasis.h*
gsSmoothPatches/gsCompositeSingleBasis.h*->gsMSplines/gsMappedSingleBasis.h*
gsMapUtils/gsMapper.h*                   ->gsMSplines/gsWeightMapper.h*


src/gsMSplines/gsMappedSpline.h
```diff
@@ -0,0 +1,110 @@
...
+    gsGeometry<T> * exportPatch(int i,gsMatrix<T> const & localCoef) const
+    { return m_compBasis->exportPatch(i,localCoef); }
+
+// Data members
+protected:
+    gsMappedBasis<d,T> * m_compBasis;
+
+}; // class gsMappedSpline
+
+} // namespace gismo
```
filiatra, line 110: at the end of this file it is missing
#ifndef GISMO_BUILD_LIB
#include GISMO_HPP_HEADER(gsMappedSpline.hpp)
#endif

this is requred for building without creating shared library (header only mode)

src/gsMSplines/gsWeightMapper.h
```diff
@@ -0,0 +1,486 @@
...
+
+#pragma once
+
+#include <gsCore/gsExport.h>
+#include <gsCore/gsLinearAlgebra.h>
+
+namespace gismo
+{
+
+class GISMO_EXPORT gsWeightMapper
```
filiatra: please add 
template<typename T>

Then for templated clases GISMO_EXPORT should be removed.

src/gsMSplines/gsWeightMapper.h
```diff
@@ -0,0 +1,486 @@
...
+{
+public:
+    typedef real_t weightType;
```
filiatra: typedef T weightType

src/gsMSplines/gsWeightMapper.cpp
```diff
@@ -0,0 +1,210 @@
+/** @file gsWeightMapper.cpp
```
filiatra: The class should be templated instead of using real_t hardcoded.
This also implies to rename this file as gsWeighMapper.hpp
and also add a new instantiation file gsWeighMapper_.cpp

src/gsMSplines/gsWeightMapper.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    Author(s): A. Bressan, F. Buchegger
+*/
+
+#include <gsMSplines/gsWeightMapper.h>
+
+namespace gismo
+{
+
+void gsWeightMapper::mapToTargetCoefs(gsMatrix<weightType> const & sourceCoefs,gsMatrix<weightType> & targetCoefs) const
```
filiatra: template<typename T>
..and similarly for all members

filiatra (changes requested):

src/gsMSplines/gsMappedSpline.h
hverhelst: gsMappedSpline.hpp does not exist but gsMappedSpline.cpp does. Should we change this?

hverhelst: Never mind; got the answer below

filiatra (approved):

PR #416, opened 2021-01-25
evalAllDers_into in gsGeometry

felixfeliz:
Added function evalAllDers_into to gsGeometry (in the same way as for deriv_into and deriv2_into).

Without this, calling evallAllDers_into on a gsGeometry leads to the implementation in gsFunctionSet, which is only implemented up to second derivatives.


filiatra (approved):
Thnx Felix!
I also added a dummy function in gsConstantFunction.
If tests are OK you can merge this

PR #422, opened 2021-02-08
Add a IETI framework

stefantakacs:
This IETI framework aims to be as general as possible.

It is based on 3 classes that represent the main ideas: gsIetiSystem, gsPrimalSystem and gsScaledDirichletPrec, which do not have any particular knowledge about the underlying discretization. This information is produced by the fourth class gsIetiMapper.

This PR is open for review, but it should not be merged until
- [x]  it was checked that it is also usable for systems (like Stokes),
- [x]  it was checked that it also provides useful tools for a dG setting,
- [x]  it is possible to also use egde averages and face averages as primal dofs
- [x]  it is possible to eliminate the corner primal dofs
- [x] the docs for gsPrimalSystem are rewritten (after now being able to eliminate the primal dofs); + some cleanup in code
- [x]  it is checked that proper memory allocation has taken place for all sparse data structures 
- [x] the algorithms are checked for correctness (e.g., by comparing the results with the results from unsupported and/or unit tests).
- [x] We should check what happens if one feeds a gsDofMapper with several variables (for a system) into the gsIetiMapper. **Now, there is an assertion that prevents the case**. The doc clearly says that there should be one mapper per variable.


examples/ieti2_example.cpp
```diff
@@ -0,0 +1,334 @@
+/** @file ieti2_example.cpp
```
stefantakacs, line 1: This is an example that shows how to use the ieti framework with the old-style assemblers.

It is not necessary to merge this into stable.

filiatra: its not bad to have it.

it would be nice for both examples to have a doumentation similarly to eg.
https://gismo.github.io/poisson2_example.html
see inside the doc folder for the corresponding file.

src/gsIeti/gsIetiMapper.h
```diff
@@ -0,0 +1,176 @@
...
+private:
+    const gsMultiBasis<T>*                        m_multiBasis;
+    gsDofMapper                                   m_dofMapperGlobal;
+    std::vector<gsDofMapper>                      m_dofMapperLocal;
+    std::vector<Matrix>                           m_fixedPart;
+    std::vector<JumpMatrix>                       m_jumpMatrices;
+    index_t                                       m_nPrimalDofs;
+    std::vector< std::vector<SparseVector> >      m_primalConstraints;
+    std::vector< std::vector<index_t> >           m_primalDofIndices;
+    unsigned                                      m_status;
```
filiatra: would be nice to describe each member with a one-liner doxy-comment here as well.

src/gsIeti/gsIetiMapper.hpp
```diff
@@ -0,0 +1,292 @@
...
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Takacs
+*/
+
+#pragma once
+
+#include <gsIeti/gsIetiMapper.h>
+#include <gsCore/gsDofMapper.h>
+#include <gsNurbs/gsKnotVector.h>
```
filiatra: are all the include files needed?

src/gsIeti/gsPrimalSystem.h
```diff
@@ -0,0 +1,205 @@
...
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Takacs
+*/
+
+#pragma once
+
+#include <gsSolver/gsLinearOperator.h>
+#include <gsSolver/gsMatrixOp.h>
+#include <gsMatrix/gsVector.h>
```
filiatra, line 17: probably some inclusions can be removed?

src/gsIeti/gsPrimalSystem.hpp
```diff
@@ -0,0 +1,167 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Takacs
+*/
+
+#pragma once
+
+#include <gsIeti/gsPrimalSystem.h>
```
filiatra: Please remove the .h file from its hpp counterpart...
Should go inside the _.cpp instead

src/gsIeti/gsIetiSystem.hpp
```diff
@@ -0,0 +1,124 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Takacs
+*/
+
+#pragma once
+
+#include <gsIeti/gsIetiSystem.h>
```
filiatra: Please remove the .h file from its hpp counterpart...
Should go inside the _.cpp instead (in some cases this can avoid cyclic includes)

src/gsIeti/gsScaledDirichletPrec.hpp
```diff
@@ -0,0 +1,217 @@
...
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): S. Takacs
+*/
+
+#pragma once
+
+#include <gsIeti/gsScaledDirichletPrec.h>
```
filiatra: Please remove the .h file from its hpp counterpart...
Should go inside the _.cpp instead

filiatra (changes requested):

filiatra:
DG in expression assembler is not in stable, will will drop C++98 in the next release, before adding it

examples/ieti2_example.cpp
stefantakacs: Agreed.

However, I have now changed the file to use the saddle point formulation (which would be necessary for inexact versions of ieti).

stefantakacs: I will fix the other issues later.

stefantakacs:
Looking forward to it, particularly also dropping C++98.

Then we should have also a look onto function signatures, like gsAssembler<>::setSparseSystem(gsSparseSystem<T> & sys)

src/gsIeti/gsIetiMapper.hpp
stefantakacs: Removed unused

src/gsIeti/gsPrimalSystem.h
stefantakacs: Removed unused

src/gsIeti/gsIetiMapper.h
stefantakacs: Comments included

examples/ieti2_example.cpp
stefantakacs: Doc in doc folder missing.

filiatra (approved):
OK. You may still consider adding some nice doxygen page, possibly with snippets, similarly to
doc/poisson2_example.dox

PR #427, opened 2021-02-10
Allow to split NURBS

stefantakacs:
This implements a function that belongs to the interface prescribed by the basis class gsGeometry. The approach delegates the work the spline case (having the effect that data has to be copied a few times).


stefantakacs:
Does not yet work as intended.

PR #428, opened 2021-02-10
This is a very special feature that could be removed from stable

stefantakacs:
This is not in use for the new IETI implementation. There is a PR in unsupported to fix the old IETI code.


filiatra (approved):
fully agree!

PR #429, opened 2021-02-11
Allow to uniformly split NURBS geometries

stefantakacs:
This implements a feature that is part of the public interface of gsGeometry and it has been around for spline geometries since quite some time (and is very helpful there).

It can be seen in action, e.g., by calling

`bin/multiGrid_example -g planar/square_with_disk.xml --SplitPatches 1 --plot`

I am not very sure if this is the most efficient way to do it (or if it could be done more directly).

### DIFF
--- a/src/gsNurbs/gsTensorNurbs.h
+++ b/src/gsNurbs/gsTensorNurbs.h
@@ -381,32 +381,21 @@ public:
         gsRationalBasis< gsTensorBSplineBasis<d,T> > * basis
             = dynamic_cast< gsRationalBasis< gsTensorBSplineBasis<d,T> > * >(m_basis);
         GISMO_ENSURE( basis, "gsTensorNurbs::uniformSplit: Dynamic cast failed." );
-        gsMatrix<T> coefs(m_coefs);
-        const size_t dm = coefs.cols(), N = coefs.rows();
-        coefs.conservativeResize(Eigen::NoChange,dm+1);
-        for (size_t i=0; i<N; ++i)
-        {
-            coefs(i,dm) = 1;
-            coefs.row(i) *= basis->weights()(i,0);
-        }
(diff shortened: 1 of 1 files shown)


filiatra:
I think it can be simplified using the functions in rationalbasis, similarly to eg. :

    // switch to n+1 dim
    gsMatrix<T> coefs1 = projectiveCoefs(coefs, weights);

   // ...do the work

    // Back to n-dim control points
    setFromProjectiveCoefs(coefs1, coefs, weights);

similarly to other functions in rationalbasis...

stefantakacs:
Thanks, I will have a look into this.

Now that there have one working version, I can create a test case as well.

src/gsNurbs/gsTensorNurbs.h
```diff
@@ -369,7 +369,38 @@ class gsTensorNurbs : public gsGeoTraits<d,T>::GeometryBase
...
+    /// Splits the geometry either two parts in direction \a dir, or if \a dir = -1
+    /// in 2^d parts, by calling splitAt() for each direction.
+    /// The function automatically searches for the midpoint the corresponding knot vector.
+    std::vector<gsGeometry<T>*> uniformSplit(index_t dir = -1) const
+    {
+        // We use the simple fact that a NURBS function in R^d is the central probjection
+        // of a spline function in R^{d+1} into R^d.
+        //
+        // Create a B-spline in R^{d+1} and split it
+        gsRationalBasis< gsTensorBSplineBasis<d,T> > * basis
```
filiatra: I think this can be replaced as
Basis & basis = this->basis(); //Basis is here gsTensorNurbsBasis

filiatra: The GISMO_ENSURE that follows can be omitted: in this context the basis is of the correct type..

filiatra (approved):

PR #431, opened 2021-02-13
Doxygen improvements

stefantakacs:


filiatra:
@stefantakacs 
Please also add ctime header in multigrid example.
Even if std::time works on some systems, in others ctime is not included automatically and compilation fails, eg.
https://cdash-ci.inria.fr/viewBuildError.php?buildid=66133

filiatra (approved):

PR #432, opened 2021-02-15
It seems that this has been a bug in gsDofMapper

stefantakacs:
I have seen this to be a problem, when testing unsupported/examples/stokesMultiPatch.cpp

@filiatra : Can you confirm that this is a proper fix?


filiatra (approved):
Yes certainly there was a "+1" missing there

PR #433, opened 2021-02-15
MG should provide basic support for systems

stefantakacs:
Now, one can choose the component index from the boundary conditions object that should be chosen. So far, always the component 0 was used.

Moreover, the options and the boundary conditions do not need to be stored as members.

### DIFF
--- a/src/gsCore/gsMultiBasis.h
+++ b/src/gsCore/gsMultiBasis.h
@@ -346,7 +346,7 @@ public:
     ///
     /// For computing the transfer matrix (but not for refinement), the \a boundaryConditions and
     /// the \a assemblerOptions have to be provided. By deault, the boundary conditions for
-    /// component 0 are chosen. Use the parameter comp to choose another one.
+    /// unknown 0 are chosen. Use the parameter unk to choose another one.
     ///
     /// \sa gsMultiBasis::uniformRefine
     void uniformRefine_withTransfer(
@@ -355,7 +355,7 @@ public:
         const gsOptionList& assemblerOptions,
         int numKnots = 1,
         int mul = 1,
-        index_t comp = 0
+        index_t unk = 0
         );
 
     /// @brief Refine the component \a comp of every basis uniformly
@@ -413,7 +413,7 @@ public:
     ///
     /// For computing the transfer matrix (but not for refinement), the \a boundaryConditions and
     /// the \a assemblerOptions have to be provided. By deault, the boundary conditions for
-    /// component 0 are chosen. Use the parameter comp to choose another one.
+    /// unknown 0 are chosen. Use the parameter unk to choose another one.
     ///
     /// \sa gsMultiBasis::uniformCoarsen
     void uniformCoarsen_withTransfer(
@@ -421,7 +421,7 @@ public:
         const gsBoundaryConditions<T>& boundaryConditions,
         const gsOptionList& assemblerOptions,
         int numKnots = 1,
-        index_t comp = 0
+        index_t unk = 0
         );
 
     /// @brief Returns the basis that corresponds to the component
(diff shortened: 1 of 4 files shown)


src/gsCore/gsMultiBasis.h
```diff
@@ -345,15 +345,17 @@ class gsMultiBasis : public gsFunctionSet<T>
...
     ///
     /// \sa gsMultiBasis::uniformRefine
     void uniformRefine_withTransfer(
         gsSparseMatrix<T, RowMajor>& transfer,
         const gsBoundaryConditions<T>& boundaryConditions,
         const gsOptionList& assemblerOptions,
         int numKnots = 1,
-        int mul = 1
+        int mul = 1,
+        index_t comp = 0
```
filiatra: As in:
void gsMultiBasis::getMapper(...,int unk,..)

it should be called "unk" (unknown) instead of component..
Terminology: For a system with eg. 2 unknowns u,p, each unknown can have dimension bigger than 1, ie. unknown u has components 0,1...,d-1

filiatra (changes requested):

filiatra:
There is still a problem in the gDofMapper constructor:

    gsDofMapper(
        const gsMultiBasis<T>         &bases,
        const gsBoundaryConditions<T> &dirichlet,
        int unk = 0)

Here there is an argument missing, which should describe the dimension (how many  components) the unknown has. eg:
index_t nComp = 1

otherwis there should be a warning in the docs that unknowns are considered scalar here..

filiatra:
OR: could it be an alternative to remove the function getMapper(...) from gsMultiBasis and do the work in place where it is used?

stefantakacs:
Isn't this a problem with the dofmapper itself?

As far as I understood, every unk has its own basis, while each component of a unk uses the same basis. So, it would make sense to allow in gsMultiBasis.refine_withTransfer to choose the unk, but to treat all components at once, i.e., we get a big transfer matrix for all components.

For this purpose, it would be required to set up a dof mapper:
a) either one for the whole unknown, or
b) one for each of the components.

However, neither gsMultiBasis::getMapper, nor gsDofMapper::init seem to support this.

Certainly, there is a member
`template <typename T>`
`void init(const gsMultiBasis<T> & bases, index_t nComp = 1);`
but this does not take boundary conditions at all.

filiatra:
> Isn't this a problem with the dofmapper itself?
it's a problem of the specific constructor of gsDofMapper..
 
> So, it would make sense to allow in gsMultiBasis.refine_withTransfer to choose the unk, but to treat all components at once, i.e., we get a big transfer matrix for all components.

exactly. However in the the mentioned constructor (with boundary conditions)  it is assumed that a single component exists.

> For this purpose, it would be required to set up a dof mapper:
> a) either one for the whole unknown, or
> b) one for each of the components.
> However, neither gsMultiBasis::getMapper, nor gsDofMapper::init seem to support this.

exactly. Option (a) is the one implemented.This is why I propose to add the nComp argument so that they will both support this (the underlying implementation is there, just the argument passing is missing)

> Certainly, there is a member
> `template <typename T>`
> `void init(const gsMultiBasis<T> & bases, index_t nComp = 1);`
> but this does not take boundary conditions at all.

nevertheless the passing of the boundary conditions to the gsDofMapper in general can ba avoided. it is better to constructo your mapper as you want it. So I propose at least to remove all BC from dofmapper, and incorporate the code inside the caller function gsMultiBasis::getMapper

stefantakacs:
I'll have a look into this. However, it will take some time.

So, you suggest that these things should be removed from gsDofMapper or should they only marked deprecated?

filiatra:
> I'll have a look into this. However, it will take some time.
> 
> So, you suggest that these things should be removed from gsDofMapper or should they only marked deprecated?

We can merge it as it is for now, and maybe deprecate the getMapper constructor or even better open an issue as deprecated just to keep track.
In any case this the problem was there before and not introduced here

filiatra (approved):

PR #434, opened 2021-02-19
Added make function for QR solver

jarlesog:
Added make function for QR solver similar to the one for LU.


filiatra (approved):

stefantakacs:
Could you please use `\relates gsSolverOp` instead of `\ingroup Solver`. This shows the make functions together with the class.

jarlesog:
I have changed it. Is it ok now?

stefantakacs:
Perfect!

PR #443, opened 2021-03-11
bugfix to gsDofMapper

jarlesog:
I believe that there was a bug here which is now fixed. 

This will create some problems for gsStokesAssembler in the unsupported repo. However, this is anyway doing something very wrong


filiatra (approved):
yes looks correct.

PR #444, opened 2021-03-19
Added a constructer for the Poisson assembler

jarlesog:
Added a constructor for the Poisson assembler which allows changing the way Dirichlet values are calculated without having to use gsOption


jarlesog:
Compiling problems. I will look into it

filiatra:
Hi,
I do not like this constructor, it is better to use the options.
Having less constructors without a bunch of options is the reason of having the options..

PR #445, opened 2021-03-21
add function to split patches at knots based on multiplicity

filiatra:


src/gsCore/gsGeometry.h
```diff
@@ -506,22 +506,30 @@ class gsGeometry : public gsFunction<T>
     }
 
     /// Embeds coefficients in \a N dimension
-    void embed(index_t N)
-    { 
+    void embed(index_t N, bool lr = true)
```
stefantakacs: What does lr mean? Maybe add to doxygen comment.

src/gsUtils/gsMesh/gsVertex.h
```diff
@@ -14,7 +14,7 @@
 #pragma once
 
 #include <gsUtils/gsMesh/gsMeshElement.h>
-#include <gsCore/gsLinearAlgebra.h>
+//#include <gsCore/gsLinearAlgebra.h>
```
stefantakacs: I think we should not comment such things out but just remove them. The same holds for the things below.

src/gsIO/gsWriteParaview.hpp
```diff
@@ -489,6 +489,12 @@ void writeSingleGeometry(gsFunction<T> const& func,
 
         if ( n == 1 )
         {
+            if (d==3)
```
stefantakacs, line 492: This one I do not understand.

filiatra: In here we have n==1, so we plot a scalar function on its domain.
When the domain is less than 3, the graph is shown in 3D.
The new if is for a scalar function on a 3D domain: the scalar field will be shown as color over the 3D domain.

stefantakacs: Thanks

stefantakacs (approved):

PR #446, opened 2021-03-24
Extend fast diagonalization

rschneckenleitner:
Allow also to handle an averaging term in fast diagonalization.

Small fix in mg.


stefantakacs (commented):
I think that this is an important improvement.

@filiatra Should I approve it, or do you want to do it?

src/gsSolver/gsPatchPreconditionersCreator.hpp
```diff
@@ -311,6 +314,14 @@ typename gsPatchPreconditionersCreator<T>::OpUPtr gsPatchPreconditionersCreator<
         // Q^T M Q = I, or M = Q^{-T} Q^{-1}
         // Q^T K Q = D, or K = Q^{-T} D Q^{-1}
 
+        if (gamma != 0)
```
filiatra: better
if ( gamma != T(0) )

filiatra (approved):

PR #452, opened 2021-05-18
Update multiGrid_example.cpp

jarlesog:
fixed so the corrected value of sweeps are given
See issue 451

@stefantakacs 
It seams that the option MG.Damping does not do anything. 
Should I remove it?


stefantakacs:
If you try
`bin/multiGrid_example --MG.Damping 3 --MG.Smoother Jacobi`
and
`bin/multiGrid_example --MG.Damping .3 --MG.Smoother Jacobi`
you get different results.
The same if you use `--MG.smoother scms`. Only Gauss-Seidel ignores it.

stefantakacs (approved):
This fixes the issue.

PR #453, opened 2021-05-25
Improve gsRemapInterface

stefantakacs:
* Make a clear distinction between affine mapping, i.e., if G_2 \circ G_1^{-1} is affine-linear, and between non-matching, i.e., the interface is not a whole edge, side, etc.
* A fitting curve is only required for the non-affine case. This case is still restricted to 2D.
* For the affine case, we can now handle any dimension since here the pre-image of the interface to the parameter domain is always an axes-parallel box.
* Simplify and clean code.
* No hidden method-parameters - all tolerances can now be configured


src/gsAssembler/gsRemapInterface.h
```diff
@@ -1,132 +1,185 @@
...
+    /// @param  checkAffine        The number of interior points (per direction) in point grid used to
+    ///                            check if the mapping is affine. If set to i, the grid consits of i interor
+    ///                            point and the two boundary points per direction, so \f$ (i+2)^d \f$ points.
+    ///                            For \a alwaysAffine or \a notAffine, no checks are performed.
+    /// @param  numSamplePoints    Number of sample points for computing fitting curve (if not affine)
+    /// @param  intervalsOfFittingCurve  Number of knot space in the fitting curve (if not affine)
+    /// @param  degreeOfFittingCurve     Spline degree of fitting curve (if not affine)
+    /// @param  equalityTolerance  Points are considered equal iff difference is smaller.
+    /// @param  newtonTolerance    Tolerance for Newton solvers (should be significantly smaller than
+    ///                            equalityTolerance)
```
filiatra: would be better to have a gsOptionList eg as argument here to avoid long list of parameters, or to anticipate adding/changing the parameters in the future.
The documentation can then refer to the expected options..

filiatra (approved):
Looks good.

PR #456, opened 2021-06-07
Made it possible to calculate the condition before the solver had ended

jarlesog:
Made it possible to calculate the CG condition number even if the method
has not converged/finished. This is useful if using the CG manually,
i.e, not using the solve() command. 

@stefantakacs  please have a look at this and let me know if it is ok


stefantakacs (approved):
OK

Pls use "Squash and merge" and not "Create merge commit".

PR #462, opened 2021-07-06
Improvements for gsVisitorDg and gsVisitorNitsche and others

stefantakacs:
Improvements for gsVisitorDg and gsVisitorNitsche:
 * By default, the grid size on the physical domain should be considered.
 * Make penalty coefficient freely configurable
 * Introduce additional parameters alpha and beta to make other DG-types (like NIPG) possible

Improvements for gsGenericAssembler:
 * Allow to assemble all terms of interest separately
 * Allow to suppress refreshing, which allows successive assembling
 * Allow to provide custom gsDofMapper

Follow-up changes for assemblers:
 * Provide default options from gsVisitorDg and gsVisitorNitsche, if applicable

Moreover:
 * Improve doxygen (make visitors visible and put them now consistently to assembler module)
 * Prove hpp-file now for all assemblers (so there should be no header-dependence of examples on visitors, which should improve compilation times.)

### DIFF
--- a/src/gsAssembler/gsAssembler.h
+++ b/src/gsAssembler/gsAssembler.h
@@ -489,6 +489,7 @@ public: /* Element visitors */
         }
     }
 
+
 public:  /* Dirichlet degrees of freedom computation */
 
     /// @brief Triggers computation of the Dirichlet dofs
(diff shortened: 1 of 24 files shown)


filiatra (approved):
Nice improvements

PR #468, opened 2021-08-31
Add gsTHBSplineBasis::active_into

arungoooo112:
The member function active_into is added into gsTHBSplineBasis,
which now can really output the non-zero reasults, Instead of simply
Inherited from class gsHTensorBasis.

### DIFF
--- a/src/gsHSplines/gsTHBSplineBasis.h
+++ b/src/gsHSplines/gsTHBSplineBasis.h
@@ -128,6 +128,9 @@ public:
     BoundaryBasisType * basisSlice(index_t dir_fixed,T par ) const;
 
     // Look at gsBasis class for documentation
+    void active_into(const gsMatrix<T>& u, gsMatrix<index_t>& result) const;
+
+    // Look at gsBasis class for documentation
     void deriv2_into(const gsMatrix<T>& u, gsMatrix<T>& result)const;
 
     // Look at gsBasis class for documentatation
--- a/src/gsHSplines/gsTHBSplineBasis.hpp
+++ b/src/gsHSplines/gsTHBSplineBasis.hpp
@@ -1058,6 +1058,81 @@ void gsTHBSplineBasis<d,T>::globalRefinement(const gsMatrix<T> & thbCoefs,
 
 
 template<short_t d, class T>
+void gsTHBSplineBasis<d,T>::active_into(const gsMatrix<T>& u, gsMatrix<index_t>& result) const
+{
+    gsMatrix<T> currPoint;
+    point low, upp, cur;
+    const int maxLevel = this->m_tree.getMaxInsLevel();
+
+    std::vector<std::vector<index_t> > temp_output;//collects the outputs
+    temp_output.resize( u.cols() );
+    size_t sz = 0;
+
+    for(index_t p = 0; p < u.cols(); p++) //for all input points
+    {
+        currPoint = u.col(p); 
+        for(short_t i = 0; i != d; ++i)
+            low[i] = m_bases[maxLevel]->knots(i).uFind( currPoint(i,0) ).uIndex();
+
+        // Identify the level of the point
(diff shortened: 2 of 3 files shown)


src/gsHSplines/gsTHBSplineBasis.h
```diff
@@ -127,6 +127,9 @@ class gsTHBSplineBasis : public gsHTensorBasis<d,T>
     /// @brief Gives back the basis at a slice in \a dir_fixed at \a par
     BoundaryBasisType * basisSlice(index_t dir_fixed,T par ) const;
 
+    // Look at gsBasis class for documentation
+    void active_into(const gsMatrix<T>& u, gsMatrix<index_t>& result) const;
```
filiatra, line 131: To have a correct over-ride of the function, in the gsHTensorBasis.h file the gsHTensorBasis::active_into function should be declared as virtual.

src/gsHSplines/gsTHBSplineBasis.hpp
```diff
@@ -1057,6 +1057,81 @@ void gsTHBSplineBasis<d,T>::globalRefinement(const gsMatrix<T> & thbCoefs,
...
+                    {
+                        temp_output[p].push_back(act);
+                    }
+                    else 
+                    {
+                        const gsSparseVector<T>& coefs = getCoefs(act);
+                        const gsTensorBSplineBasis<d, T>& base =
+                            *this->m_bases[this->m_is_truncated[act]];
+
+                        gsMatrix<index_t> ind;
```
filiatra: It is more efficient to declare all gsMatrix<..> objects outside all loops.
This way the memory is allocated once and reused throught the loop - it can make a significant changes in speed.

src/gsHSplines/gsTHBSplineBasis.hpp
```diff
@@ -1057,6 +1057,81 @@ void gsTHBSplineBasis<d,T>::globalRefinement(const gsMatrix<T> & thbCoefs,
...
+                        base.active_into(currPoint, ind);
+
+                        for (index_t k = 0; k < ind.rows(); ++k) 
+                        {
+                            if (!gsClose(coefs(ind.at(k)), 0.0, 1e-10))
```
filiatra: We may use different arithmetic, with different precision characteristics.
Instead of 1e-10, you can try to use  std::numeric_limits < T > ::epsilon()
which will set a machine zero according to the arithmetic used

filiatra (changes requested):

src/gsHSplines/gsTHBSplineBasis.hpp
arungoooo112: Thanks for your advice. I've learned.

src/gsHSplines/gsTHBSplineBasis.hpp
arungoooo112: Yes, It is my fault. And now I fixed this problem.
: )

filiatra:
Thnx for the updates! The function looks already great.
Another idea of improvement could be the following:
If I am not mistaken, the coefficient vector "coefs", which is sparse vector, only contains explicit zeros for inactive coefficients.
If it is the case,  then simply 
          if ( 0 == coefs(ind.at(k)) )
would work as a check..
This is something to check.
Other than that, I will accept the PR after some checks..

arungoooo112:
Thank you for your patience and advice!
This change makes the code simpler and easier to understand.
And, I really need to pay attention to more details.
: )

filiatra (approved):

stefantakacs:
@arungoooo112 : Please merge this commit if ready.

arungoooo112:
@stefantakacs Merging is blocked for me. I am not authorized to merge this pull request.

arungoooo112 (commented):
review changes

stefantakacs:
Since it has been approved by @filiatra, I think I can merge this as well.

PR #469, opened 2021-09-03
Add reading of Iges files

filiatra:


src/gsIO/gsFileData.hpp
```diff
@@ -1535,31 +1536,620 @@ bool gsFileData<T>::readObjFile( String const & fn )
...
+    E_TYPE= 0,
+    PD_PTR= 1,
+    PD_CNT= 2,
+    ELAYER= 3
+};
+
+int entity[MAXENTITY][4];
+int entity_sum=0,         /* A number of entities in array entity[] */
+    emark=0;              /* for cycling over entity[] */
+
```
stefantakacs: Does this really need global vars?

filiatra: fixed, thnx.

PR #473, opened 2021-10-04
Adding p-multigrid method 

roeltielen24:
Provides a new example file with a p and hp multigrid solver.

Since any p-refinement yields non-nested spaces, there are several options for the restriction and prolongation matrices.


stefantakacs:
Hi, we should discuss these days how to base this better on existing multigrid code.

stefantakacs:
@filiatra : Could you have a look on this PR?

stefantakacs (approved):

PR #508, opened 2022-01-13
Update gsMultiGrid.h

mmoelle1:
Fixed warnings about override virtual functions without using the `override` keyword.


stefantakacs:
Thanks for the warning fix, but c++ requires that it must be `const override`, never `override const`.

stefantakacs (approved):

PR #510, opened 2022-01-21
Reset the submodule download system avoiding git-submodule

filiatra:


mmoelle1:
I always wanted to rename `gsUnum` and the flag `UNUM` into `gsUniversal` and `UNIVERSAL`. This might be the right time to do it. Can you do this, please?

filiatra:
> I always wanted to rename `gsUnum` and the flag `UNUM` into `gsUniversal` and `UNIVERSAL`. This might be the right time to do it. Can you do this, please?

OK I will

mmoelle1 (commented):
Could you please rename exprtk_unum_posit into exprtk_universal_posit?

PR #512, opened 2022-01-28
Hotfix integral bdr

hverhelst:
Add integralBdrBc in the expression evaluator to be able to evaluate boundary integrals defined using boundary conditions


hverhelst:
This PR is not based on a recent version of stable; hence it should be omitted

PR #535, opened 2022-05-16
Submodule gs unstructured splines

hverhelst:
Adds submodules and updates gsMappedSplines/gsMappedBasis. 
Compatible with the (to be made public) gsUnstructuredSplines module

-----

Pull requests can only be merged after at least one review (and
approval) from @gismo/admins.

Code submitted to the stable branch should be clean, well-documented
and free of bugs.

# Please consider the following checklist before issuing a pull request:
- [ ] Have you added an explanation of what your changes do and why
  you'd like us to include them?
- [ ] Have you documented any new codes using Doxygen comments?
- [ ] Have you written new tests or examples for your changes?
-----

PR #544, opened 2022-09-05
reverting inheritance the way it was when gsMappedSpline was still gsCompositeGeom

dominik-mokris-mtu:
As far as I can tell, the stable class ```gsMappedSpline``` has been made from the unsupported class ```gsCompositeGeom```. However, there is an important difference:

- ```gsCompositeGeom``` inherits from ```gsGeoTraits<d,T>::GeometryBase```
- ```gsMappedSpline``` inherits from ```gsFunctionSet<T>```

Unless there is a reason for this change (@hverhelst or @filiatra probably know more), I would like to revert it and make ```gsMappedSpline``` inherit from ```gsGeoTraits<d,T>::GeometryBase```. In my view, this is required so that I can bring ```gsCompositeIncrSmoothnessGeom``` and the rest of ```gsSmoothPatches``` into a branch of stable.

The magic with ```GISMO_BASIS_ACCESSORS``` has been necessary to get around the rather cryptic compiler errors.

approval) from @gismo/admins.

and free of bugs.

# Le checklist
**Yes, see above**
**No, I don't think it is necessary in this case.**
**No, I don't think it is necessary in this case.**
-----

### DIFF
--- a/src/gsMSplines/gsMappedSpline.h
+++ b/src/gsMSplines/gsMappedSpline.h
@@ -26,7 +26,7 @@ namespace gismo
 template<short_t d,class T> class gsMappedSingleSpline;
 
 template<short_t d,class T>
-class gsMappedSpline : public gsFunctionSet<T>
+class gsMappedSpline : public gsGeoTraits<d,T>::GeometryBase
 {
     friend class gsMappedSingleSpline<d,T>;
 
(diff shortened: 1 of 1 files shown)


hverhelst:
Hey @dominik-mokris-mtu ,

We have a PR #535  in progress that makes some modifications on the mapped-splines and their integration into different assemblers. In that PR and in the previous introduction of `gsMSplines`, the class `gsCompositeIncrSmoothnessGeom` not needed anymore since `gsMappedSingleSpline` inherits from `gsFunction`. Does that make sense? Or is there a specific reason why one would need `gsCompositeIncrSmoothnessGeom` instead of `gsMappedSingleSpline` and `gsMappedSpline` for geometry evaluations?

dominik-mokris-mtu:
Oh. Upon closer inspection, we don't seem to be using ```gsCompositeIncrSmoothnessGeom``` either. Maybe I should have noticed that before spending a few hours trying to fix it.^^

Thanks for the clarification, I am looking forward to #535.

PR #547, opened 2022-10-07
Contact

ckarampa:
approval) from @gismo/admins.

and free of bugs.

request:
-----

This branch has the additions / modifications to gismo that needed to be made in order to enable some basic contact simulations. 

Some main changes are:
- Introduced the `gsCPPInterface` ( i.e. Closest Point Projection ), needed to associate points from one side of the interface with their closest neighbor on the other. In doing so, the `gsAffineFunction` has been **replaced** in `gsExprEvaluator<T>::computeInterface_impl()` thus prior functionality has certainly been affected. **TODO** We need to come up with a way to facilitate computations with either approach, without affecting current functionality.
- Since the projected points are not guaranteed to lie on a single element, the use of the evaluator flag `SAME_ELEMENT` needed to be restricted only to where it is truly the case.
- Introduced the `flipSide` assembler/evaluator option, which swaps the roles of ~~master~~ `right()` and ~~slave~~ `left()`
- Added `gsExprEvaluator::evalIfc` to evaluate expressions on interfaces.
- Created `assembleJacobian` in an effort to "automatically" assemble the tangent stiffness matrix given a residual as an expression, by means of central differences.
- When the measure of a geometryMap is requested on one of its boundaries, the measure of the outer normal vector is returned instead.

Changes under `gsExpressions`:
- Added expressions for the positive and negative part of an expression ( aka element-wise ramp functions )
- Created `gsFeSolution::setComponent` to easily set a component of the solution vector equal to some predefined value, on all or on a specific patch.
- Introduced `gsGeometryMap::deformBy( gsFeSolution )` to deform the geometry map so that we can can track the current state.
- Created a `gsMatrix` x `expr` multiplication, but also an expression that represents fixedMatrices( e.g. an Id matrix or a matrix of ones) , so I guess one of the two is redundant. **TODO** Decide which solution should be accepted.

Last but not least:
- [ ] **TODO** Add more Doxygen documentation.


stefantakacs:
Hi, what's the difference between gsCPPInterface and gsRemapInterface ?

ckarampa:
Hello, I have not used `gsRemapInterface` but from what I see in the documentation, there you assume that the two sides of the interface share a (more or less) common edge. For `gsCPPInterface` this assumption is not made and the points from one side are mapped to their closest neighbour (in the physical space) on the opposite side (even if they are far apart).

ckarampa:
Regarding the last point on my original post, about the multiplication of expressions by gsMatrices, after a discussion with Angelos, the later solution was preferred. It has been implemented in [8bfb91f](https://github.com/gismo/gismo/pull/547/commits/8bfb91f2096bf4158e679c03ec60cba40e2d3b5a).

filiatra (approved):

PR #572, opened 2022-12-19
Paraview multiple fields

ckarampa:
approval) from @gismo/admins.

and free of bugs.

request:
- [x] Have you written new tests or examples for your changes?
-----

## What has been implemented
As discussed in the G+Smo Developer Meeting ( 12 / 12 / 2022 ) I have implemented a 'new' way to export ParaView files that allows :
- The export of multiple solution fields
- The export of multiple time steps
- Arbitrary export path
- User specified number of evaluation points, plot precision
- Optional plotting of elements / control net.

All this functionality is shared between the new `gsParaviewDataSet` class and the existing `gsParaviewCollection` which has been heavily modified (although from a user's perspective not much has changed). This new "framework" currently supports `gsExpressions` and `gsFields`, but if there is interest it can be expanded to other types of data.

The `poisson2_example` and `heatEquation_example` have been updated to showcase the features introduced in this PR.

## Questions:
- [x] It is unclear to me what the `part` attribute of the `<DataSet>` xml tag does, and if it necessary at all. If anyone knows and uses that, please explain and we then shall keep the option to specify it. Otherwise I suggest we remove this functionality.


src/gsAssembler/gsExprEvaluator.h
```diff
@@ -324,13 +325,19 @@ class gsExprEvaluator
                        std::string const & fn)
     { writeParaview_impl<E,false>(expr,m_exprdata->getMap(),fn); }
 
+    template<class E>
+    std::vector<std::string> expr2vtk(const expr::_expr<E> & expr, std::string label="SolutionField",unsigned nPts=1000, unsigned precision=5);
+
+    std::vector<std::string> geoMap2vtk(const geometryMap geoMap, unsigned nPts=1000, unsigned precision=5);
```
filiatra: This funtion geoMap2vtk is never used, is it needed ?

filiatra (changes requested):

src/gsIO/gsParaviewDataSet.cpp
```diff
@@ -0,0 +1,108 @@
...
+            file <<"</PointData>\n\n\n<!-- GEOMETRY -->\n<Points>\n";
+            file << points[k];
+            file << "</Points>\n</Piece>\n</StructuredGrid>\n</VTKFile>";
+            file.close();
+            if (plotControlNet)
+            {
+                writeSingleControlNet( m_geometry->piece(k), m_basename + "_cnet" + std::to_string(k));
+                m_filenames.push_back( m_basename + "_cnet" + std::to_string(k)+".vtp");
+            } 
+            if ( plotElements)
```
filiatra, line 88: Here: there should be all the code from the gsWriteParaview function, in particular the computation of the resolution for the plotting of elements

src/gsAssembler/gsExprEvaluator.h
ckarampa: Indeed since you suggested to store a `gsMultiPatch` as a member of `gsParaviewDataSet` and not a `gsGeometryMap` this function is not used anywhere, so I guess it should be removed altogether

src/gsIO/gsParaviewDataSet.cpp
ckarampa: I don't quite get what you are suggesting, could you elaborate ?

ckarampa: Implemented in [60f9deb](https://github.com/gismo/gismo/pull/572/commits/60f9deb05b79554f9487160d62320abcfc18b343) if you want to have a look

filiatra (approved):

PR #577, opened 2023-01-19
New OptimizeForArchitecture macro

mmoelle1:
The new OptimizeForArchitecture macro is more robust as it tests the CPU-specific compiler flags during the CMake configuration one by one to make sure that all of them compile properly with the selected compiler and in concert with the other flags. This slows down the CMake configuration step but allows for much deeper optimization.


filiatra (approved):
It's good. I like the message warning "it can take some time", it makes it friendly to the user.

PR #579, opened 2023-02-01
removing references to deleted files from gsInstall.cmake

dominik-mokris-mtu:
This is a tiny extension of #577: the deleted cmake files should not be referenced in the install scripts.


dominik-mokris-mtu:
The issue has been resolved in de18aed54805743c19f949ec4705f6e14b912298.

PR #599, opened 2023-04-14
Fix DirichletStrategy in poisson2 example

felixfeliz:
The poisson2 example did not converge when the exact solution in `poisson2d_bvp.xml` is changed to a function that is non-zero at the Dirichlet boundary (e.g. the function from the file +1). This is because DirichletStrategy was set to 11, but `gsExprAssembler` expects 1 for elimination and defaults to ignoring the dofs. Thus, I changed it to 1 in `poisson2d_bvp.xml` together with the description taken from `gsExprAssembler<T>::defaultOptions()`.

More generally, I noticed that `gsExprAssembler` expects DirichletStrategy to be 0,1,2, while `gsAssembler` expects 0,11,12,13 as defined in `dirichlet::strategy`. Maybe it would make sense to unify this.

-----

### DIFF
--- a/filedata/pde/poisson2d_bvp.xml
+++ b/filedata/pde/poisson2d_bvp.xml
@@ -43,7 +43,7 @@
 
   <!-- Assembler options -->
  <OptionList id="4">
-  <int label="DirichletStrategy" desc="Method for enforcement of Dirichlet BCs [11..14]" value="11"/>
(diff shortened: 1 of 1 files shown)


filiatra (approved):

filiatra:
Indeed we should unify the label numbers

PR #606, opened 2023-04-24
Unify indices of Dirichlet strategy in gsExprAssembler and gsAssembler

felixfeliz:
Currently, in the options of `gsExprAssembler` the value of `DirichletStrategy` corresponding to elimination of the dofs is `1`, while `gsAssembler` uses the values defined in `dirichlet` and therefore expects `dirichlet::elimination=11` for elimination.
This change uses `dirichlet::elimination` instead of `1` also in gsExprAssembler, in order to unify the two usages.
I searched the stable branch for all instances of DirichletStrategy and the old value 1 was not used anywhere except for the default options of `gsExprAssembler` and the poisson2 example. 

HOWEVER, this change would break people's private code in a way that is hard to debug if they explicitly set DirichletStrategy to be 1 instead of using the default options (It took me quite a while to understand this problem in my code). So I am not 100% sure how to proceed.

-----

### DIFF
--- a/filedata/pde/poisson2d_bvp.xml
+++ b/filedata/pde/poisson2d_bvp.xml
@@ -43,7 +43,7 @@
 
   <!-- Assembler options -->
  <OptionList id="4">
-  <int label="DirichletStrategy" desc="Strategy related to enforcement of Dirichlet DoF values [0ignore, 1:eliminate, 2:..]" value="1"/>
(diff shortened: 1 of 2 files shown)


filiatra (approved):
The "1" was introduced very recently, and it does create problems with the existing coding.
So I also think it is better to revert to existing coding, since less (private) code needs to be updated..

PR #625, opened 2023-07-05
Arbitrary Level Hierarchical Splines

hverhelst:
Improves the support of the `gsHTensorBasis<d,T>::addLevel`. The code can be tested with the new `gsHTensorBasis<d,T>::testPartitionOfUnity`

request:
-----


src/gsMSplines/gsMappedBasis.hpp
```diff
@@ -82,36 +82,36 @@ void gsMappedBasis<d,T>::addLocalIndicesOfPatchSide(const patchSide& ps,index_t
...
-        patchSide firstPs = iter->first();
-        addLocalIndicesOfPatchSide(firstPs,offset,locals);
-        patchSide secondPs = iter->second();
-        addLocalIndicesOfPatchSide(secondPs,offset,locals);
-    }
-    sort( locals.begin(), locals.end() );
-    locals.erase( unique( locals.begin(), locals.end() ), locals.end() );
-    m_mapper->sourceToTarget(locals,indices);
-}
+// template<short_t d,class T>
```
filiatra: why where these commented? please add a short comment

filiatra (approved):

src/gsMSplines/gsMappedBasis.hpp
hverhelst: Good question. I think I had the hypothesis that they were wrong. Now I uncommented them back and the code still works. Though, I added a comment that when used, they should be checked more carefully.

PR #630, opened 2023-07-26
Added a MinRes-QLP implementation

jarlesog:
An implementation of the MinRes-QLP method

I used both the matlab code and petsc code in my implementation.
It works and seems to be more stable then normal minres
example run:
./bin/linearSolvers_example -n 5000

There are still some possible extensions to do (written as todo's in the comments) like:
1) Create accessors for values of interest, like, Arnorm, xnorm etc.
2) Add the same flaging system the matlab code uses
3) Merge the minres and minresQLP code like petsc and matlab has

But for now the code is operational and the extensions are not needed to solve a linear system,
they just provide additional functionality.

-----

### DIFF
--- a/examples/linearSolvers_example.cpp
+++ b/examples/linearSolvers_example.cpp
@@ -15,7 +15,7 @@
 
 #include <iostream>
 #include <gismo.h>
-
+#include <gsSolver/gsMinResQLP.h>
 
 using namespace gismo;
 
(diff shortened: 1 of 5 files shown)


examples/linearSolvers_example.cpp
```diff
@@ -15,7 +15,7 @@
 
 #include <iostream>
 #include <gismo.h>
-
+#include <gsSolver/gsMinResQLP.h>
```
filiatra: please add the include in the file src/gismo.h instead.

filiatra (changes requested):

examples/linearSolvers_example.cpp
jarlesog: Done

filiatra (approved):

PR #236, opened 2018-10-14
Add binary format for XML i/O

filiatra:
NEW: base64 data encoding for XML I/O

PR #634, opened 2023-08-31
Base64 I/O for Paraview and XML

jzwar:
# Motivation
Although I was aware of the gismo-PR #236, I saw that there has not been an update since 2018 and it also relies on an external library. Further, it only acted on the XML-files without support for Paraview exports.

The need to redo the implementation of this feature originated as we had fields with very small magnitudes, which is very tedious with fixed-point string conversion.

# Implementation
The implementation is based on [this](https://github.com/tataratat/bezman/blob/main/src/utils/base64.hpp), which I implemented as an interface to @j042's python library [splinepy](github.com/tatatarat/splinepy). It relied on `c++17` features, so adaptations were required, and further overloads for `gsMatrix` were added. The `ToVTK`-function was updated accordingly.

## Details
- [x] Moved functions related to `GoTools` into a separate file, although it seems to be used nowhere (?)
- [x] Added a new class of static functions only for base64 encoding and decoding
- [x] Added `base64` option to Paraview collection and added argument to `toVTK`
- [x] Updated some variable names to make code clearer, formatted modified sections 
- [x] XML I/O

# Complementary
I would add the following, should there be a demand for it:
- [x] Support KnotVectors in XML input
- [ ] Tests for De/Encoding (they already exist, but I did not know where to put them)


mmoelle1:
I think that adding base64 support for the knot vectors would be very good. We are planning to implement something similar to the `Geometry` tag to write solutions as native spline objects in XML. These solutions will often have many coefficients and hence benefit largely from base64 support.

jzwar:
> I think that adding base64 support for the knot vectors would be very good. We are planning to implement something similar to the `Geometry` tag to write solutions as native spline objects in XML. These solutions will often have many coefficients and hence benefit largely from base64 support.

Done and rebased

filiatra:
Very usefull !
Tests can go in a new file inside /unittests
Please set in cmake GISMO_BUILD_UNITTESTS=ON
to be able to compile (make unitttestss) and run (./bin/unittests) the unittests.

jzwar:
@filiatra As discussed, I looked into the structure date is currently written to file. Adding a flag to gsFileData will not be sufficient. Streaming a file into `gsFileData` calls the `put` function, which is non-member. So, as far as I can see, no trivial way to add format flags to the output of the children without changing the layout of every put function.

filiatra:
OK then let's go on with this one as it is and let the writing for later.

src/gsIO/gsXml.hpp
```diff
@@ -66,24 +67,36 @@ char * makeValue(const gsMatrix<T> & value, gsXmlTree & data,
...
+template <class T>
+void getMatrixFromXml(gsXmlNode* node, unsigned const& rows,
+                      unsigned const& cols, gsMatrix<T>& result,
+                      const std::string& base_type_flag) {
+    // Make sure that flag is in lower case for comparisons
+    std::string base_type_flag_;
+    base_type_flag_.reserve(base_type_flag.size());
+    std::transform(base_type_flag.cbegin(), base_type_flag.cend(),
+                   std::back_inserter(base_type_flag_),
+                   [](unsigned char c) { return std::tolower(c); });
```
filiatra, line 79: Please use "::tolower" instead of "std::tolower" because some windows compiles like it better.

src/gsNurbs/gsKnotVector.hpp
```diff
@@ -65,12 +68,21 @@ class gsXml< gsKnotVector<T> >
...
-        str.str( node->value() );
-        for (T knot; gsGetReal(str, knot);)
-            knotValues.push_back(knot);
-
+        // Read knots from list
+        gsXmlAttribute* format = node->first_attribute("format");
+        std::string format_flag = format ? format->value() : "ascii";
+        std::transform(format_flag.cbegin(), format_flag.cend(),
+                       format_flag.begin(),
+                       [](unsigned char c) { return std::tolower(c); });
```
filiatra, line 76: Please use "::tolower" instead of "std::tolower" because some windows compiles like it better.

filiatra (changes requested):

PR #635, opened 2023-08-31
Keep Sparsity Pattern (Mini-PR)

jzwar:
# Motivation
When doing multiple assemblies (e.g. in a non-linear problem, or in an optimization), accessing the sparse matrix can be a bottleneck when using the expression assembler. Whenever the `clearMatrix()` command was called, the matrix was reinitialized and the sparsity pattern was deleted.

In this suggestion, the matrix is multiplied by zero whenever clearMatrix is called to avoid re-initialization. This is also made default behavior.

With OpenMP in my optimization, it was significantly faster.

### DIFF
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -303,6 +303,12 @@ public:
 
     void clearRhs() { m_rhs.setZero(); }
 
+    /**
+     * @brief Re-Init Matrix (set zero by default)
+     *
+     * @param save_sparsety_pattern only modify values but keep sparsety
+     * information by multiplying matrix by zero in-place
+     */
     void clearMatrix(const bool& save_sparsety_pattern = true) {
       if (save_sparsety_pattern) {
         m_matrix *= 0.0;
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -303,24 +303,29 @@ public:
 
     void clearRhs() { m_rhs.setZero(); }
 
-    void clearMatrix()
-    {
-        m_matrix = gsSparseMatrix<T>(numTestDofs(), numDofs());
+    void clearMatrix(const bool& save_sparsety_pattern = false) {
+        if (save_sparsety_pattern) {
+            m_matrix *= 0.0;
+        } else {
+            m_matrix = gsSparseMatrix<T>(numTestDofs(), numDofs());
 
-        if ( 0 == m_matrix.rows() || 0 == m_matrix.cols() )
-            gsWarn << " No internal DOFs, zero sized system.\n";
-        else
-        {
-            // Pick up values from options
-            const T bdA       = m_options.getReal("bdA");
-            const index_t bdB = m_options.getInt("bdB");
-            const T bdO       = m_options.getReal("bdO");
-            T nz = 1;
-            const short_t dim = m_exprdata->multiBasis().domainDim();
-            for (short_t i = 0; i != dim; ++i)
-                nz *= bdA * static_cast<T>(m_exprdata->multiBasis().maxDegree(i)) + static_cast<T>(bdB);
+            if (0 == m_matrix.rows() || 0 == m_matrix.cols())
+              gsWarn << " No internal DOFs, zero sized system.\n";
+            else {
+              // Pick up values from options
+              const T bdA = m_options.getReal("bdA");
+              const index_t bdB = m_options.getInt("bdB");
+              const T bdO = m_options.getReal("bdO");
+              T nz = 1;
+              const short_t dim = m_exprdata->multiBasis().domainDim();
+              for (short_t i = 0; i != dim; ++i)
+                nz *= bdA * static_cast<T>(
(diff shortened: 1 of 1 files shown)


weinmueller:
Hello, that is interesting. Good to know. Could you try it with `setZero()` instead of `*=0`? Maybe this is faster (and cleaner)?

jzwar:
> Hello, that is interesting. Good to know. Could you try it with `setZero()` instead of `*=0`? Maybe this is faster (and cleaner)?

Hi :) The function `setZero`, as implemented [here](https://github.com/gismo/gismo/blob/8475faa0d6b9affd5d06988adf8c18a75b4566c4/external/gsEigen/src/SparseCore/SparseMatrix.h#L253), will not reallocate memory, but (as I understand it) set the index list to zero, basically "forgetting" that a value is stored at said location. If the function is called, subsequent access to an arbitrary value in the matrix will result in a call of the insert function. That's why imo, it is not suited for this application. 

An alternative approach to multiplication by zero, would be:
``` c++
std::fill(m_matrix.valuePtr(),
                  m_matrix.valuePtr() + m_matrix.nonZeros(), 0.);
```

weinmueller:
> If the function is called, subsequent access to an arbitrary value in the matrix will result in a call of the insert function. That's why imo, it is not suited for this application.

Okay, so if I understand correctly, you want to keep the memory of the sparse matrix? And then you modify somewhere the matrix with coefRef() which leads to the fact that this is faster?

src/gsAssembler/gsExprAssembler.h
```diff
@@ -298,30 +298,40 @@ class gsExprAssembler
...
-            const T bdA       = m_options.getReal("bdA");
-            const index_t bdB = m_options.getInt("bdB");
-            const T bdO       = m_options.getReal("bdO");
-            T nz = 1;
-            const short_t dim = m_exprdata->multiBasis().domainDim();
-            for (short_t i = 0; i != dim; ++i)
-                nz *= bdA * static_cast<T>(m_exprdata->multiBasis().maxDegree(i)) + static_cast<T>(bdB);
-
-            m_matrix.reservePerColumn(numBlocks()*cast<T,index_t>(nz*(1.0+bdO)) );
+        if (0 == m_matrix.rows() || 0 == m_matrix.cols())
```
filiatra: Please use 4-space identation

filiatra (changes requested):

jzwar:
> > If the function is called, subsequent access to an arbitrary value in the matrix will result in a call of the insert function. That's why imo, it is not suited for this application.
> 
> Okay, so if I understand correctly, you want to keep the memory of the sparse matrix? And then you modify somewhere the matrix with coefRef() which leads to the fact that this is faster?

@weinmueller Not only the memory, but also the non-zero locations, i.e., the sparsity information. Basically, the indices are kept as matrix entries with the value `0`. 

If you call coeffRef() for a non-existing matrix entry, then `insert` will be called. From the [eigen documentation](https://eigen.tuxfamily.org/dox/classEigen_1_1SparseMatrix.html#title22):
>Assuming memory has been appropriately reserved, this function performs a sorted insertion in O(1) if the elements of each inner vector are inserted in increasing inner index order, and in O(nnz_j) for a random insertion. 

As increasing inner-index order is not assured, that can be an expensive task (when I tested, this was most important for many-patch-geometries). Adding to that: When compiling with OpenMP, `insert` blocks the thread to ensure safe access to the matrix, which slows down the assembly significantly.

mmoelle1:
If you allow me to comment on this. @filiatra and I started with some improvement of the assemblers in which we do exactly what @jzwar needs. First generate the sparsity pattern (sequentially) and then do the assembly (with OpenMP parallelisation and `atomic` updates). For that we implemented some special version of the `coefRef()` that throws an error of the position is not available. The normal `coefRef()` would insert it, I guess. It's not yet in stable but will be soon, I hope.

jzwar:
> If you allow me to comment on this. @filiatra and I started with some improvement of the assemblers in which we do exactly what @jzwar needs. First generate the sparsity pattern (sequentially) and then do the assembly (with OpenMP parallelisation and `atomic` updates). For that we implemented some special version of the `coefRef()` that throws an error of the position is not available. The normal `coefRef()` would insert it, I guess. It's not yet in stable but will be soon, I hope.

@mmoelle1 I have been playing around with `setFromTriplets` for the same reason, which also helped to improve assembly times by quite a bit. Do you already have a plan to store the sparsity information? In that case, this PR would be redundant and I could close it. Otherwise, would you prefer the `std::fill` approach or rather `*=0.`?

filiatra (approved):

filiatra:
Please add a commit description using the keyword IMPROVED before mergins, eg:

IMPROVED: avoid re-initialization of .....


Regarding the comment of Matthias, I think that this PR is combatible with what we want to achieve, so in my opinion it can be mearged

jzwar:
> Please add a commit description using the keyword IMPROVED before mergins, eg:
> 
> IMPROVED: avoid re-initialization of .....
> 
> Regarding the comment of Matthias, I think that this PR is combatible with what we want to achieve, so in my opinion it can be mearged

@filiatra I am not authorized to merge myself, the merge button is greyed out.

PR #642, opened 2023-09-28
Checking if the Eigen symlink already exists before creating a new one

nilskohl:
A second build configuration from the same source directory makes cmake complain and exit otherwise.

> CMake Error: failed to create symbolic link '.../gismo/external/Eigen': File exists

The commit just wraps what was introduced in f3e05c394f2828ac8d9c1eeaec473193755db4a1

New to gismo - therefore happy to discuss this :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,9 @@ endif()
 ## #################################################################
 
 #Fix for Eigen directory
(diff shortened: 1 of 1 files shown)


filiatra (approved):
Hi nilskohl, that's a good fix, thanx !

PR #662, opened 2023-11-15
Floating point precision showcase, quadrature lookup guard

nilskohl:
Hi,

this PR adds:

* minor fixes for MPFR,
* conditionals to refuse quadrature rule lookup if the underlying floating point precision is higher than that of the precomputed points and values,
* a showcase (example) that illustrates the issues of using insufficient floating point precision with respect to discretization-error-accuracy (this example has also been used to test the quadrature weight lookup fix described in the previous item - even with MPFR and wide mantissas the quadrature implementation defaulted to the lookup tables if it was not manually forced to recompute the values leading to an inaccurate system).

The guard for the quadrature lookup is not necessarily the perfect solution, but it is simple and should just work the same way as if a precomputed quadrature rule could not be found due to other reasons. So let me know if you have a better fix in mind :)

### DIFF
--- a/examples/precision_example.cpp
+++ b/examples/precision_example.cpp
@@ -12,7 +12,6 @@
 */
 
 #include <gismo.h>
-#include <gsAssembler/gsPoissonAssembler.h>
 
 using namespace gismo;
 
--- a/examples/precision_example.cpp
+++ b/examples/precision_example.cpp
@@ -127,7 +127,7 @@ int main(int argc, char *argv[])
 
     int defaultDegree = 5;
     dataIn.degree = defaultDegree;
-    int maxRefine = 9;
+    int maxRefine = 7;
 
     // Setting default to IEEE 754 "double"
     int mantissa = 53;
--- a/examples/precision_example.cpp
+++ b/examples/precision_example.cpp
@@ -1,0 +1,222 @@
+/** @file precision_example.cpp
+
+    @brief Example illustrating the effect of quantization and rounding errors in the absence of sufficient precision.
+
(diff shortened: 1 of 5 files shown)


examples/precision_example.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): N. Kohl
+*/
+
+#include <gismo.h>
+#include <gsAssembler/gsPoissonAssembler.h>
```
filiatra: gsPoissonAssembler.h is already included inside gismo.h, please remove it form here.

filiatra (changes requested):

examples/precision_example.cpp
nilskohl: Fixed in [1933fd1](https://github.com/gismo/gismo/pull/662/commits/1933fd1ef02e805950ad6d8166304e512b5ae796).

filiatra (approved):

nilskohl:
Just checking in to ask if this can be merged :)

filiatra:
> Just checking in to ask if this can be merged :)

Hi,
yes it can! As soon as it's approved (which it is right now) you can push the button !

nilskohl:
Great! But I am not authorized to do that unfortunately 😄

filiatra:
strange, OK I do it this time.

PR #670, opened 2023-12-16
Small changes in gsExprAssembler and gsExprEvaluator

hverhelst:
gsExprAssembler.h: Activate try-catch only when not in debug mode
gsExprEvaluator.h: paraview function compiler error

NEW: 
IMPROVED: 
Allow try-catch statement in gsExprAssembler only outside debug mode, to make debugging easier
FIXED: 
Compiler error for write paraview in gsExprEvaluator
API:


src/gsAssembler/gsExprEvaluator.h
```diff
@@ -18,7 +18,7 @@
 #include <gsAssembler/gsQuadrature.h>
 #include <gsAssembler/gsRemapInterface.h>
 #include <gsAssembler/gsCPPInterface.h>
-//#include <gsIO/gsWriteParaview.h>
+#include <gsIO/gsWriteParaview.h>
```
filiatra: this include creates a problem in header-only builds.
Can we do without it?

filiatra (changes requested):

src/gsAssembler/gsExprEvaluator.h
hverhelst: When using `writeParaview` of `gsExprEvaluator`, it gives a compiler error:
```
.../src/gsAssembler/gsExprEvaluator.h:929:34: error: ‘gsWriteParaviewTPgrid’ was not declared in this scope
  929 |             gsWriteParaviewTPgrid( gmap ? pts : pt.toMatrix(), // parameters
      |             ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  930 |                                   vals,
      |                                   ~~~~~
  931 |                                   pt.numPointsCwise(), fileName );
```
So it needs to find `gsWriteParaviewTPgrid`.

filiatra: Hm, I checked again on gcc and it still works. If it does not, instead of including gsWriteParaview.h we should just forward declare gsWriteParaviewTPgrid

filiatra: OK, in 30e6e4e175273190a868721cced618d70fe4b5c6 I move the forward declarations, it is not perfect but it should fix the issue.

filiatra (approved):

Issue #672, opened 2023-12-20
THB: knot indices can overflow


dominik-mokris-mtu:
I am planning the following change. If you have any related thoughts, please let me know.

### Problem
When fitting a certain big point cloud that I am not allowed to share with a THB-surface, I started with a rather fine basis of level 0 with 127 x 15 internal knots. Fitting itself worked well but I ran into trouble when converting the resulting THB-surface to a B-spline using `gsTHBSpline<d, T>::convertToBSpline(gsTensorBSpline<d,T>& result)`: the resulting surface was severely mangled. Confusingly, using my older G+Smo installation from 2020-03-02 worked well.

After quite some debugging, the problem turned out to be one of the many changes done in #334, in particular in 53eeeb3. There the function `gsHDomain<d,T>::local2globalIndex` has been changed to use `index_t` instead of `unsigned`. On my machine, this leads to an overflow and thus to a big _negative_ value being written to `result[i]`. Instead of crashing right away, G+Smo attempts to work with this box with negative indices, leading to strange results.

### Proposed solution
After a discussion with @filiatra today, my plan is to make G+Smo (or, more precisely, `gsHTensorBasis`) to set `m_indexLevel` depending on the number of knots of the coarsest basis and the type of `index_t` to the highest value possible without overflowing.

A low-cost alternative would be to just warn the user that they are about to overflow and ask them to recompile with a bigger type used for `index_t`.

While doing so, I intend to do a little clean-up of the files in question.

### Progress
There is a dedicated branch [issue-672](https://github.com/gismo/gismo/tree/issue-672), where I try to stick to atomary commits so as not to drown in all the changes.

- [X] `gsHDomain` should consistently use its template `T`, which I will rename to `Z` in order to indicate that it is an integer type,
- [X] `kdnode` will get renamed to `gsKdNode` to be consistent with the usual G+Smo logic,
- [X] `gsAabb` will get renamed to `gsAABB` to make it clear that it is an abbreviation,
- [X] `gsVSegment` will get separated into its own file in order to be easier to locate,
- [X] commented out code will be removed,
- [X] documentation will get checked and updated if necessary,
- [X] `m_indexLevel` will be computed adaptively as suggested above,
- [X] a unit test should be added to prevent the issue from reappearing,
- [X]  fix #677 by correcting `gsTHBSpline::convertToBSpline`.

dominik-mokris-mtu:
[...]

Issue #677, opened 2024-01-04
THB boxes are indexed in a level much higher than m_indexLevel


dominik-mokris-mtu:
**Update:** the problem turned out to be something different, see my comment below.

While working on #672, I noticed the following peculiar behaviour.

---

**Describe the bug**
The domain of a THB-spline is represented using an `m_tree` of the type `gsHDomain`. This consists of many boxes, each of them storing its level and two corners. In order to prevent rounding problems, the corners are not real numbers but indices of the knots. However, one needs to decide which level of the hierarchy to use for the indexing.

From what's written in gsHDomain.h, one would think that they are indexed in terms of `m_indexLevel`:
```cpp
    /// The level of the box representation (global indices)
    unsigned m_indexLevel;
```

However, when working with the boxes, the knots get recomputed to the global level _once again_, thus using a much higher level. See the example in the following section. 

---
**To Reproduce**
We need **two steps**: a printout in gsHDomain.hpp and a test file.

In gsHDomain.hpp find the function `insertBox` and print the `iBox` between `local2globalIndex` and the error check as follows
```cpp
    local2globalIndex( iBox.first , static_cast<unsigned>(lvl), iBox.first );
    local2globalIndex( iBox.second, static_cast<unsigned>(lvl), iBox.second);

    gsInfo << "iBox: ("
           << iBox.first[0] << ", " << iBox.first[1] << ") x ("
           << iBox.second[0] << ", " << iBox.second[1] << ")" << std::endl;

    // Ensure that the box is within the valid limits
    if ( ( iBox.first.array() >= m_upperIndex.array() ).any() )
```

Now create a new cpp file in the `examples` folder.
```cpp
int main()
{
    gsInfo << "First, we read the face geometry." << std::endl;
    gsFileData<> in("surfaces/thbs_face_3levels.xml");
    gsTHBSpline<2> thb;
    in.getId<gsTHBSpline<2>>(0, thb);

    gsInfo << std::endl;
    gsInfo << "Second, we analyze the gsHDomain." << std::endl;
    gsInfo << "m_indexLevel:   " << thb.basis().tree().getIndexLevel() << std::endl;
    gsInfo << "m_maxInsLevel:  " << thb.basis().tree().getMaxInsLevel() << std::endl;
    gsInfo << "upper corner:   ("
           << thb.basis().tree().upperCorner()[0] << ", "
           << thb.basis().tree().upperCorner()[1] << ")" << std::endl;
[...]

PR #679, opened 2024-01-09
Issues #672 and #677.

dominik-mokris-mtu:
approval) from @gismo/admins.

and free of bugs.

IMPROVED: `gsTHBSpline::convertToBSpline` is now not restricted to `d=2`,
IMPROVED: source code clean-up in the gsHSplines folder,
FIXED: issues #672 and #677.

For each new/improved/fixed/api change use a new line with the
respective word prefix in capital.

request:
  Yes, cf. issues #672 and #677.
- [X] Have you documented any new codes using Doxygen comments? Not really, as it is mainly modification of existing code.
- [X] Have you written new tests or examples for your changes? Yes, there is a new unittest `gsFixedIssues_test.cpp`, which we can extend in future when fixing other issues.
-----

@filiatra, before merging I would like your opinion on a few things:
- `gsTHBSpline::convertToBSpline` now does not explicitly call `delete`. Am I right in thinking that the unique pointer will take care of what `makeGeometry` produced?
- In `gsHDomain::insertBox` we check the box but only partially. Would it make sense to be more strict and check that both its corners are within the limits?
- I think the unittest can be made more precise but I don't see an elegant way. Any ideas?
- There are still several `unsigned`s around, especially when counting the levels. Would it make sense to make the level a `short_d`?

### DIFF
--- a/src/gsHSplines/gsAABB.h
+++ b/src/gsHSplines/gsAABB.h
@@ -1,6 +1,7 @@
-/** @file gsKdNode.h
+/** @file gsAABB.h
 
-    @brief Provides declaration of the tree node.
+    @brief Declarations of the class gsAABB, i.e., the axis-aligned
+    bounding box.
 
     This file is part of the G+Smo library.
     
--- a/src/gsHSplines/gsAABB.h
+++ b/src/gsHSplines/gsAABB.h
@@ -1,0 +1,106 @@
+/** @file gsKdNode.h
+
+    @brief Provides declaration of the tree node.
+
+    This file is part of the G+Smo library.
+    
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+    Author(s): A. Mantzaflaris, D. Mokris
+*/
+
+# pragma once
+
+namespace gismo {
+
+/**
+    @brief Struct of for an Axis-aligned bounding box
+
+    Template parameters
+    \param d is the dimension
+    \param Z is the box-coordinate index type
+    
+    \ingroup HSplines
+*/
+
+template<short_t d, class Z = index_t>
+struct gsAABB
+{
+public:
+    typedef gsVector<Z, d> point;
+
+    gsAABB(const point & low, const point & upp, index_t lvl)
+    :
+    first(low), second(upp), level(lvl)
+    { }
+
+    gsAABB(const point & low, const point & upp)
+    :
+    gsAABB(low, upp, -1)
+    { }
+
+    gsAABB(const point & upp)
+    :
+    second(upp), level(-1)
+    {
+        first.setZero();
+    }
+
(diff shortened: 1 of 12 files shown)


src/gsHSplines/gsAABB.h
```diff
@@ -0,0 +1,106 @@
+/** @file gsKdNode.h
```
filiatra: please update @file and @brief

src/gsHSplines/gsHBox.h
```diff
@@ -14,6 +14,8 @@
 #pragma once
 
 #include <gsIO/gsXml.h>
+#include <gsCore/gsLinearAlgebra.h>
+#include <gsHSplines/gsAABB.h>
```
filiatra: are the new includes needed? 
It is better to minimize inclusions if somehow the compiler finds the definition without them, or even add them to the hpp file instead.
(same consideration for all the new includes in the PR)

filiatra (changes requested):

src/gsHSplines/gsAABB.h
dominik-mokris-mtu: Done in [cedfd10](https://github.com/gismo/gismo/pull/679/commits/cedfd1091be19649cea5f2248df3b62465b699ab).

src/gsHSplines/gsHBox.h
dominik-mokris-mtu: My reasoning was that it is better to include everything that is used in the file. Otherwise, I cannot guarantee that everything will be fine on the computer of someone else. And I didn't like the prospect of someone getting an error in a code that they did not touch, which is what can theoretically happen.

With respect to these two particular includes:
- In `gsHBox.h` we need only the declaration of `gsAABB`, its definition is only needed in `gsHBox.hpp`. If you prefer, I could put a forward declaration in `gsHBox.h` and move the include to `gsHBox.hpp`. That way we should be safe while not including too much.
- I included `gsLinearAlgebra.h` because at some point after touching the order of includes, I got a compilation error complaining about `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` on line 55. That macro is defined somewhere in the linear algebra, hence the include.

filiatra: I agree with the reasoning to include what is used, but minimal inclusions help for better compilation.. Indeed a better way is to forward-declare when possible.
So yes please move the include to gsHBox.hpp
Since linear algebra is needed OK to keep it!

filiatra (approved):

src/gsHSplines/gsHBox.h
dominik-mokris-mtu: Great! I will go through the includes that I touched with this suggestion in mind.

dominik-mokris-mtu: Done in [f8edbc](https://github.com/gismo/gismo/pull/679/commits/f8edbc069f2bb0d22fd7a069c40c473205f97e1c), I hope I haven't forgotten anything.

PR #686, opened 2024-02-04
Fix compile error in header-only mode and real_t differ from `double`

mmoelle1:
This PR fixes a missing hpp-file include when compiled in header-only mode and a problem when compiled with `real_t` different from `double`.


filiatra (approved):
Thnx!

PR #691, opened 2024-02-23
Enable Python bindings based on cppyy

felixfeliz:
# Summary
NEW: This adds new Python bindings based on the Python-C++ bindings generator [cppyy](https://cppyy.readthedocs.io/en/latest/)
NEW: Add two examples for their usage

# Why
The main difference to the existing Pybind11 bindings is that cppyy is based on just in time compilation and the bindings are generated at runtime. For G+Smo this is useful, since it enables us to freely use the `gsExprAssembler` in Python. This was not possible with pre-compiled bindings, since `gsExprAssembler` is based on templated expressions that are not known when G+Smo is compiled. With the cppyy bindings, we can in principle define any bilinear form also in Python. Another side effect is that bindings do not need to be defined manually and all definitions from G+Smo can be immediately used in Python.

# Configuration
Configure gismo with the options `-DGISMO_WITH_CPPYY=On -DCMAKE_CXX_STANDARD=20`. This creates a Python package `gismo_cppyy` in the `build/cppyy` directory. Build `gismo` as usual.
`make wheel` generates the pip wheel in `cppyy/dist` and `make install-bindings` installs it using pip.

The package looks for the G+Smo sources in the source directory and for the shared library in the build directory.

# How to use
G+Smo can then be loaded in Python using `from gismo_cppyy import gismo`.
See `python_examples/poisson_example.py` and `python_examples/fitting_example.py` for examples of the usage.

Numpy arrays and gsMatrix objects can be converted in Python using the method `tonumpy()` and the class method `fromnumpy()` of `gismo.gsMatrix["real_t"]` ( or any other data type). Numpy arrays can also be converted in the same way to `gsAsMatrix`, which avoids copying.

- [x] Have you written new tests or examples for your changes?
-----


CMakeLists.txt
```diff
@@ -333,6 +333,47 @@ else()
 endif()
 add_subdirectory(doc)
 
+## #################################################################
```
filiatra, line 336: It would be better to move this cmake script in a new cppyy/CMakeLists.txt
and do 
```
if(GISMO_WITH_CPPYY)
  add_subdirectory(cppyy) 
endif()
```
here.
This way the main CMakeLists.txt is more clear and also the cppyy related script more autonomous.

cmake/gsInstall.cmake
```diff
@@ -159,6 +159,10 @@ install(FILES
 #install(EXPORT gismoTargets DESTINATION
 #  "${CMAKE_INSTALL_DIR}" COMPONENT devel)
 
+if(GISMO_WITH_CPPYY)
```
filiatra: probably the install connamd can go inside the cppyy/CMakeLists.txt

filiatra (changes requested):
Looks good!

CMakeLists.txt
felixfeliz: That is a good idea! I moved the configuration into cppyy/CMakeLists.txt

cmake/gsInstall.cmake
felixfeliz: Done

filiatra (approved):

PR #700, opened 2024-03-30
Bsplinecurve intersection

jiyess:
New feature for efficiently computing all intersections bewteen two arbitrary-dimensional B-Spline curves in `gsBSpline` class

BSpline_intersections

NEW: 
1. `T gsBSpline<>::pseudoCurvature() const` for calculating curvature approximation for a B-spline curve by dividing the total length of the polyline formed by its control points by the direct distance between the first and last control point;
2. `std::vector<internal::gsCurveIntersectionResult<T>> gsBSpline<>::intersect(const gsBSpline<T>& other, T tolerance = 1e-5) const` for computing all intersections with another B-Spline curve;
3. std::vector<gsBSpline<T>> gsBSpline<>::toBezier(T tolerance=1e-15) const for converting a BSpline curve into Bezier segments;
4. `gsCurveCurveIntersection.h` for functions and classes used for `gsBSpline::intersect`.

IMPROVED: 
none

FIXED: 

1. Fix one bug in `gsBSpline<T> gsBSpline<T>::segmentFromTo(T u0, T u1, T tolerance) const`

API:
none

request:
-----

### DIFF
--- a/examples/bSplineCurve_example.cpp
+++ b/examples/bSplineCurve_example.cpp
@@ -20,11 +20,13 @@ using namespace gismo;
 int main(int argc, char *argv[])
 {
     bool plot = false; // If set to true, paraview file is generated and launched on exit
-    bool trim = false; // If set to true, paraview file is generated and launched on exit
+    bool trim = false; // If set to true, trim/merge operations are displayed
+    bool intersect = false; // If set to true, intersection example is displayed
 
     gsCmdLine cmd("Tutorial 01 shows the use of BSpline curves.");
     cmd.addSwitch("plot", "Plot result in ParaView format", plot);
     cmd.addSwitch("trim", "Basic trim/merge operations", trim);
+    cmd.addSwitch("intersect", "Intersection operations", intersect);
 
     try { cmd.getValues(argc,argv); } catch (int rv) { return rv; }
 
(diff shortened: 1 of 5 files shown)


examples/bSplineCurve_example.cpp
```diff
@@ -78,9 +80,43 @@ int main(int argc, char *argv[])
       mergedCurve.merge(&segmentRight);
       gsInfo << "The merged curve: " << mergedCurve << "\n";
       gsWriteParaview( mergedCurve, "mergedCurve", 100);
+
+      // convert it into bezier segments
+      std::vector<gsBSpline<>> bezSegments = mergedCurve.toBezier();
```
filiatra: it would be better to use a gsMultiPatch<> object as return type for the Bezier segmetns

src/gsNurbs/gsBSpline.h
```diff
@@ -22,6 +22,7 @@
 
 #include <gsUtils/gsPointGrid.h>
 
+#include <gsNurbs/gsCurveCurveIntersection.h>
```
filiatra: Please add here a forward declaration
template<class T> gsCurveCurveIntersection;

and move the 
#include <gsNurbs/gsCurveCurveIntersection.h>
to the hpp file

src/gsNurbs/gsBSpline.hpp
```diff
@@ -127,6 +128,62 @@ gsBSpline<T> gsBSpline<T>::segmentFromTo(T u0, T u1, T tolerance) const
...
+    gsBSpline<T> crv1 = this->segmentFromTo(hull.b1.getRange().getMin(), hull.b1.getRange().getMax());
+    gsBSpline<T> crv2 = other.segmentFromTo(hull.b2.getRange().getMin(), hull.b2.getRange().getMax());
+
+    internal::gsCurveCurveDistanceSystem<T> obj(crv1, crv2);
+    gsMatrix<T, 2, 1> uv;
+    uv(0, 0) = 0.5 * (crv1.domainStart() + crv1.domainEnd());
+    uv(1, 0) = 0.5 * (crv2.domainStart() + crv2.domainEnd());
+    T distance = obj.compute(uv, tolerance);
+
+    if (distance < std::max(1e-10, tolerance)) {
```
filiatra: please use math::max instead of std::max
and math::min instead of std::min
everywhere

src/gsNurbs/gsCurveCurveIntersection.h
```diff
@@ -0,0 +1,360 @@
...
+  if (val == 0) return 0;  // collinear
+
+  return (val > 0) ? 1 : 2; // clock or counterclock wise
+}
+
+// Given three collinear points p, q, r, the function checks if
+// point q lies on the line segment 'pr'
+template<class T>
+bool onSegment(const gsPoint2d<T> &p, const gsPoint2d<T> &q, const gsPoint2d<T> &r) {
+  if (q.x() <= std::max(p.x(), r.x()) && q.x() >= std::min(p.x(), r.x()) &&
```
filiatra: please use math::max instead of std::max
and math::min instead of std::min
everywhere

filiatra (changes requested):

src/gsNurbs/gsBSpline.hpp
jiyess: Done!

src/gsNurbs/gsCurveCurveIntersection.h
jiyess: Done!

src/gsNurbs/gsBSpline.h
jiyess: A forward declaration for `gsCurveIntersectResult` is added in `gsBSpline.h`, and `#include <gsNurbs/gsCurveCurveIntersection.h>` is added into the `gsBSpline.hpp` file. However, it does not work if I remove the header file `#include <gsNurbs/gsCurveCurveIntersection.h>` from `gsBSpline.h`. So I leave it there. Do you have any suggestions?

examples/bSplineCurve_example.cpp
jiyess: Done!

filiatra (approved):

PR #707, opened 2024-05-08
fix linker error: undefined reference

AndreaBressan:
FIXED: the combineMappers templated function with real_t was not instantiated in the library causing linker errors

instantiating revealed a different error that is also fixed

FIXED: an error in the combineMappers templated function that called copyBlockTo passing a matrix instead of a gsWeightMapper

### DIFF
--- a/src/gsMSplines/gsWeightMapper_.cpp
+++ b/src/gsMSplines/gsWeightMapper_.cpp
@@ -26,9 +26,12 @@ namespace gismo
 
     //Utils
 
-    GISMO_EXPORT index_t reorderMapperTarget (gsWeightMapper<real_t> &mapper, const std::vector<index_t>& permutation, gsPermutationMatrix* permMatrix=NULL);
-    GISMO_EXPORT  gsWeightMapper<real_t>* combineMappers (const std::vector<gsWeightMapper<real_t>*> &mappers, std::vector<index_t> &shifts, bool needShifting=true);
-    GISMO_EXPORT  void combineMappers (const std::vector<gsWeightMapper<real_t>*> &mappers, bool needShifting, std::vector<index_t> &shifts, gsWeightMapper<real_t>& result);
+    TEMPLATE_INST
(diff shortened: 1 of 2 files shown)


src/gsMSplines/gsWeightMapper_.cpp
```diff
@@ -26,9 +26,12 @@ namespace gismo
 
     //Utils
 
-    GISMO_EXPORT index_t reorderMapperTarget (gsWeightMapper<real_t> &mapper, const std::vector<index_t>& permutation, gsPermutationMatrix* permMatrix=NULL);
-    GISMO_EXPORT  gsWeightMapper<real_t>* combineMappers (const std::vector<gsWeightMapper<real_t>*> &mappers, std::vector<index_t> &shifts, bool needShifting=true);
-    GISMO_EXPORT  void combineMappers (const std::vector<gsWeightMapper<real_t>*> &mappers, bool needShifting, std::vector<index_t> &shifts, gsWeightMapper<real_t>& result);
+    TEMPLATE_INST
+    index_t reorderMapperTarget (gsWeightMapper<real_t> &mapper, const std::vector<index_t>& permutation, gsPermutationMatrix* permMatrix=NULL);
```
hverhelst: Windows does not like the default argument here. Could you remove it?

src/gsMSplines/gsWeightMapper_.cpp
```diff
@@ -26,9 +26,12 @@ namespace gismo
...
+    TEMPLATE_INST
+    gsWeightMapper<real_t>* combineMappers (const std::vector<gsWeightMapper<real_t>*> &mappers, std::vector<index_t> &shifts, bool needShifting=true);
```
hverhelst: Same here, remove the default argument

hverhelst (changes requested):

src/gsMSplines/gsWeightMapper_.cpp
AndreaBressan: done

hverhelst (approved):

PR #708, opened 2024-05-13
IETI example for Stokes

stefantakacs:
New example that shows how Expression Assembler + IETI framework can be used to solve systems of PDEs.

A corresponding geometry has been added.

A comprehensive documentation has been added.

Moreover, this improves IETI plot functionality by directly constructing the field to be written out for Paraview. There is no need to have to setup a new assembler just for this purpose since the IETI mapper has all the data required to incorporate the Dirichlet conditions.


filiatra (approved):
Nice! Please merge it.

PR #710, opened 2024-05-15
Avoids compilation error.

### LINKED ISSUE #705: Compiler error.
I am getting that `dim` is a function, not a `short_t` variable. It should probably be `dim()`. Or `m_dim`...

https://github.com/gismo/gismo/blob/4d2819ad2e131cbfc527b1553635188b73cc63b4/src/gsModeling/gsTemplate.h#L588

I might be saying something silly, I've just met G+Smo a few hours ago. :-)

I have changed mine to `dim()` and it compiles, now. Looks like some sort of regression, but the sources have not been modified for a long time.

andre-caldas:
FIXED: https://github.com/gismo/gismo/issues/705

For some reason we are returning a method's name instead of calling it.

### DIFF
--- a/src/gsModeling/gsTemplate.h
+++ b/src/gsModeling/gsTemplate.h
@@ -585,7 +585,7 @@ public:
         return os; 
     }
 
-    short_t parDim() const { return dim; }
+    short_t parDim() const { return dim(); }
 
     /// Number of patches    
     index_t nPatches() const { return this->size(); }


andre-caldas:
The MACOS build is giving a warning related to this line (`cmake/ctest_script.cmake`):
```
cmake_minimum_required(VERSION 2.8.12)
```

It says that anything below 3.5 will not be compatible with new cmake versions. I do not know anything about cmake... but I suppose that understand that the line can be simply commented out or removed, although this is not the suggestion the deprecation warning gives.

Also, the tests give an error, but there is no test output to indicate what the error is. All we have is the message:
```
CMake Error at /Users/runner/work/gismo/gismo/cmake/ctest_script.cmake:623 (message):
  Some Tests failed.

Error: Process completed with exit code 255.
```

hverhelst:
@andre-caldas the errors on MacOS seem to be there also in stable, it needs another fix. FYI, if you click "details" for the test, you see the test result. Under the header "Running ctest" you will find something like:
```
CDASH LINK:
https://cdash-ci.irisa.fr/index.php?project=Gismo&date=2024-05-16&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=1219-macos-latest[710]&field2=site&compare2=65&value2=stable-a4150ba_[github]
```
That will bring you to the CDash server, where you will find more information about the compiler errors.

@filiatra , do you know anything about the MacOS compilation?

PR #712, opened 2024-05-16
Small changes for basis, knot vector, domain iterator

hverhelst:
NEW:
Paraview write for boxes represented by matrices 
Nestedness check of knot vector
Extra constructor for `gsTensorDomainBoundaryIterator` 

IMPROVED: 

FIXED:
ID of `gsTensorDomainIterator`

API:
Make `gsBasis<T>::evalAllDersSingle_into` with argument `gsMatrix<T> &` deprecated, and add `gsBasis<T>::evalAllDersSingle_into` with argument `std::vector<gsMatrix<T>>` in line with `evalAllDers_into`. Only `gsBSplineBasis::evalAllDersSingle_into` overloads, but throws a warning for a bug anyways. 

----

request:
-----


filiatra (changes requested):
There is the function:

gsBasisFun<T> function(index_t i) const

that can be used to get a handle on a single basis function.
This should do the same as "computesingle".

Can you tryto use that and if it works the same, remove computeSinge ?
it looks a bit of duplicated functionality.

hverhelst:
[PR712](https://trello.com/c/7dCbOzSw/3-pr712)

hverhelst:
> There is the function:
> 
> gsBasisFun function(index_t i) const
> 
> that can be used to get a handle on a single basis function. This should do the same as "computesingle".
> 
> Can you tryto use that and if it works the same, remove computeSinge ? it looks a bit of duplicated functionality.

I have reverted the `computeSingle`. There is still the `evalAllDersDingle_into`, but that one provides an implementation of an existing function with that name using different input arguments.

hverhelst (commented):
I have reverted the computeSingle. There is still the evalAllDersDingle_into, but that one provides an implementation of an existing function with that name using different input arguments.

hverhelst:
I have checked this PR again, @filiatra could you have a look.
One improvement would be to replace the `gsWriteParaview` function for boxes to a version where they are not written in `nBoxes` number of files, but instead just in one file (with an unstructured grid or so. I don't know yet which type to use in Paraview.

filiatra (approved):

PR #721, opened 2024-07-07
Bindings to the Optim library

hverhelst:
Add bindings to the [optim](https://github.com/kthohr/optim/) package

NEW: 
add `optional/gsOptim/`

IMPROVED:

FIXED: 

API:
`gsOptProblem::numConstraints()` returned the number of lower bounds (which are box constraints). Now it returns the member `m_numConstraints` which is a user-defined flag. This keeps it functional for non-linear constraints, but the user has to set it explicitly.

-----
request:
-----


sofimp23:
If I include the optional submodule `gsOptim` via the `ccmake` interface, I get the following linking error:

> gismo/external/optim/header_only_version/misc/optim_options.hpp:221:10: fatal error: 'BaseMatrixOps/include/BaseMatrixOps.hpp' file not found #include "BaseMatrixOps/include/BaseMatrixOps.hpp"

and indeed `BaseMatrixOps` is empty.

If I include the optional submodule `gsOptim` via editing `../submodules.txt`, everything works.

Though, `BaseMatrixOps` is not moved into `optim/include`, but copied.
optim CMakeList.txt, line 28:
`file(RENAME ${gismo_externals}/BaseMatrixOps ${gismo_externals}/optim/include/BaseMatrixOps)`
From the CMake documentation:
> file(RENAME <oldname> <newname>)
> Move a file or directory within a filesystem from <oldname> to <newname>, replacing the destination atomically.

filiatra:
It should work now, please try it again.

optional/gsOptim/gsOptim_.cpp
```diff
@@ -0,0 +1,8 @@
+#include <gsCore/gsTemplateTools.h>
```
filiatra: No need for _cpp file, since all code is templated (inside .h file).

Actually this compiled code is never called in this scenario.

filiatra (changes requested):

filiatra (approved):

CMakeLists.txt
```diff
@@ -204,6 +204,11 @@ if (";${GISMO_OPTIONAL};" MATCHES ";gsCoDiPack;")
   gismo_fetch_module(gsCoDiPack)
 endif()
 
+# WHY IS THIS NEEDED HERE????
```
filiatra: this should be removed. I think

filiatra (changes requested):

hverhelst:
There is an issue when compiling `examples/optimizer_example.cpp`
```bash
In file included from gismo/external/optim/header_only_version/optim.hpp:24,
                 from gismo/optional/gsOptim/gsOptim.h:20,
                 from gismo/examples/optimizer_example.cpp:28:
gismo/external/optim/header_only_version/misc/optim_options.hpp:221:10: fatal error: BaseMatrixOps/include/BaseMatrixOps.hpp: No such file or directory
  221 | #include "BaseMatrixOps/include/BaseMatrixOps.hpp"
```
Whereas the file `gismo/optional/gsOptim/examples/optim_test.cpp` compiles fine (and uses the file `BaseMatrixOps/include/BaseMatrixOps.hpp`).

@filiatra or @jiyess could you have a look if there is something wrong in the [`CMakeLists.txt`](https://github.com/gismo/gismo/blob/6078c588cc3cd4c9c6a962eb42a6354777b9975e/optional/gsOptim/CMakeLists.txt)?

filiatra (approved):

sofimp23:
The optimizer example does not compile for me.

`make optimizer_example` 

gives the following error:
```
gismo/external/optim/header_only_version/misc/optim_misc.hpp:25:10: fatal error: 'misc/optim_structs.hpp' file not found
#include "misc/optim_structs.hpp"
```

PR #736, opened 2024-09-05
Installs a simple pkg-config file.

### LINKED ISSUE #734: Provide a "gismo.pc" pkg-config file
When I compile and install G+Smo in my Debian GNU/Linux system, I get header files and a shared library. So, my program is supposed to link to libgismo and therefore a linker option is needed.

A portable way for a library to inform others about the needed compiler and linker flags is by [providing a pkg-config file](https://people.freedesktop.org/~dbn/pkg-config-guide.html).

By the way, I use [meason](https://mesonbuild.com/) (instead of cmake) as my build system. And to declare dependency on the G+Smo library [I would need one simple line of code if there was an installed pkg-config (or cmake "config")](https://github.com/andre-caldas/minimal_ogre/blob/ed039aeb5694b2570bcef9c90d937578f954110b/meson.build#L19).

Do you want me to provide this? I can research a little about producing pkg-config files in CMake and make a PR.

andre-caldas:
NEW: Installs a pkg-config file to be used by build systems that are not cmake.

FIXED: https://github.com/gismo/gismo/issues/734

A `gismo.pc` file is installed to `lib/pkgconfig/`.
In meson, for example, in order to compile and link to an installed G+Smo package, you just have to add `dependency('gismo')` to your list of dependencies.

### DIFF
--- a/cmake/gsInstall.cmake
+++ b/cmake/gsInstall.cmake
@@ -178,13 +178,13 @@ endif(GISMO_BUILD_LIB)
 # Install pkg-config file
 if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
   # FreeBSD uses ${PREFIX}/libdata/pkgconfig
-  set(GISMO_PKGCONFIG_INSTALLDIR "libdata/pkgconfig")
+  set(GISMO_PKGCONFIG_INSTALL_DIR "libdata/pkgconfig")
 else()
-  set(GISMO_PKGCONFIG_INSTALLDIR "${LIB_INSTALL_DIR}/pkgconfig")
+  set(GISMO_PKGCONFIG_INSTALL_DIR "${LIB_INSTALL_DIR}/pkgconfig")
 endif()
 
 install(FILES "${PROJECT_BINARY_DIR}/gismo.pc"
-  DESTINATION "${GISMO_PKGCONFIG_INSTALLDIR}/"
+  DESTINATION "${GISMO_PKGCONFIG_INSTALL_DIR}/"
   RENAME "${PROJECT_NAME}.pc")
 
 install(DIRECTORY "${PROJECT_SOURCE_DIR}/cmake/ofa"
--- a/cmake/gsInstall.cmake
+++ b/cmake/gsInstall.cmake
@@ -163,9 +163,22 @@ install(FILES
 #install(EXPORT gismoTargets DESTINATION
 #  "${CMAKE_INSTALL_DIR}" COMPONENT devel)
 
+# Produce pkg-config file
+configure_file ("${PROJECT_SOURCE_DIR}/gismo_lib.pc.in"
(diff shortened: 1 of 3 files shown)


cmake/gsInstall.cmake
```diff
@@ -163,10 +163,23 @@ install(FILES
...
+  message ("Configure with -DGISMO_BUILD_LIB=ON to compile the library")
+
+# Produce pkg-config file
+  configure_file ("${PROJECT_SOURCE_DIR}/gismo_nolib.pc.in"
+                  "${PROJECT_BINARY_DIR}/gismo.pc" @ONLY)
 endif(GISMO_BUILD_LIB)
 
+# Install pkg-config file
+install(FILES "${PROJECT_BINARY_DIR}/gismo.pc"
+  DESTINATION "lib/pkgconfig/"
```
filiatra: Thanx for this addtion!

Is this path platform independent?
Maybe we can use 

${LIB_INSTALL_DIR}/pkgconfig/

instead ?

filiatra (changes requested):

cmake/gsInstall.cmake
andre-caldas: I totally agree with you! I don't really know.

In my system, SDL, OGRE and CGAL they all put `.cmake` files in `lib/cmake/xxx/`. And they all put `.pc` files in `lib/pkgconfig/`.

But I have no idea about specifications and recommendations. I will research a little.

andre-caldas: > ${LIB_INSTALL_DIR}/pkgconfig/
> instead ?

Of course! I was a little confused because I was under the impression `LIB_INSTALL_DIR` was `lib/gismo`. Sorry! `LIB_INSTALL_DIR` is fine. Others usually use `CMAKE_INSTALL_LIBDIR`.

> Is this path platform independent? Maybe we can use

Just for the record, it seems FreeBSD does not like this very much... SDL does it like this:
https://github.com/libsdl-org/SDL/blob/26728c321fa9ba5300e90fc14f32fa1f974b9ab8/CMakeLists.txt#L3448-L3453

Do we have a way to detect FreeBSD, or shall I use `LIB_INSTALL_DIR`?

filiatra: OK, thnx for all the research.
You may detect FreeBSD as
if( ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" )

filiatra (approved):
You may detect FreeBSD as
if( ${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD" )

cmake/gsInstall.cmake
andre-caldas: Done.

I am very sorry I used to commits... First I had used a variable named `xxx_INSTALLDIR`, then I got it changed to `xxx_INSTALL_DIR`.

filiatra:
Thnx, I will just merge this.

PR #745, opened 2024-11-24
New parallel schema for gsExprAssembler - assemble function

GiannisKarampinis:
Introduce a slightly different parallel schema for assemble() of gsExprAssembler by subdividing the existing parallel region into two different parallel regions in which: a) 1st parallel region serves as the collector of quadrature weights and points (patchId is also kept) of the whole domain into an std::vector, b) 2nd parallel region calls the precompute and assembles the stiffness matrix in parallel for the whole domain with a manual splitting into chunks. Potential gains should be considered due to the sequential memory storing of elem data. To be tested in cluster infrastructure in order to capture the scalability and compare it with the existing parallel schema.

### DIFF
--- a/examples/poisson2_example.cpp
+++ b/examples/poisson2_example.cpp
@@ -159,6 +159,7 @@ int main(int argc, char *argv[])
 
         gsInfo<< "." <<std::flush;// Assemblying done
 
+/*
         timer.restart();
         solver.compute( A.matrix() );
         solVector = solver.solve(A.rhs());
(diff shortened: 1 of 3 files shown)


hverhelst:
Hi @GiannisKarampinis !
I like the idea, but will need to look a bit in it.
Just for now, since you did not evaluate on the performance yet: could you convert the PR to draft for the moment, until it is clear what the performance benefits are?
In addition, do you need anything to test the performance? If you want to discuss it, shoot me a message!

PR #740, opened 2024-09-23
Implement Bezier Element export for ParaView

ckarampa:
VTK (and subsequently ParaView) now natively supports Bezier elements. This allows us to export geometries ( and solution fields in a future PR) by means of Bezier extraction.

### NEW:
- `gsWriteParaviewBezier` accepts a `gsMultiPatch` and exports it as a *single* `.vtu` file containing `VTK_BEZIER_QUADRILATERAL` elements. Currently this is implemented only for bivariate geometries.
- `gsMultiPatch::BezierExtraction()` implements Bezier extraction via quasi interpolation and has been brought over from `gismo/gsLsDyna` and returns a multi patch consisting of all the bezier elements.. The struct `ElementBlock` and method `BezierOperator` are necessary for its  operation. The struct has been 'hidden' inside the `internal` namespace and `BezierOperator` has been marked private. 

### IMPROVED:
- Static functions of `gsParaviewDataSet` moved to a separate file, `gsParaviewUtils`. These deal mainly with formatting data to the .vtk format.
- `toDataArray` now also supports integers, matrices of integers, and custom xml tag attribute-value pairs.
- A flag has been added to `inputOutput_example`, which exports the geometry as bezier elements.
- `filedata/surfaces/teapot.xml` now contains a multi patch definition.

### TODO:
- [ ] To avoid compilation errors the PR in `gismo/gsLsDyna` which deletes the struct/funtions from there needs to be accepted along with this PR.

# Please consider the following checklist before issuing a pull
request:
- [x] Have you added an explanation of what your changes do and why
  you'd like us to include them?
- [x] Have you documented any new codes using Doxygen comments?
- [x] Have you written new tests or examples for your changes?
-----

PR #748, opened 2024-12-04
Multi-row output for gsWriteParaview 

lventavinuela:
In the old version, when a field was written with `gsWriteParaviewTPgrid` and the field had more than 3 components, the file would export all the components in the field but would assign `NumberOfComponents=3` (hard coded). This gave wrong results in Paraview.

In this PR with export fields with more than 3 rows a Tensors and write the number of components as the rows of the field (`eval_field`). We have tested this within gsElasticity for fields with 6 (Voigt notation) and 9 components (tensor), giving correct behavior. For fields with 3 components, Vector type is used in Paraview.

-----

NEW: 
IMPROVED: 
FIXED:
Support export of fields with more than 3 components (e.g. full stress tensor in 3D) to Paraview
API:

-----
request:
- [ ] Have you added an explanation of what your changes do and why
-----


hverhelst (approved):

hverhelst:
@filiatra and I checked why the GitLab CI fails  and it is due to the fact that the tests take around 90 minutes (let's say, 90+/- 2 minutes), so sometimes they succeed (if test time < 90 minites), other times they fail due to time out. We'll fix this in another PR.

CC @mmoelle1

ckarampa:
Hello all, if you check out #740 (not merged yet) there is a file called `gsParaviewUtils` and there we have some functions that convert gsMatrices  to the raw xml string to be written. It might be a good idea to start using that instead of hardcoding raw xml into each program, because it is more maintainable that way.

filiatra (approved):
Looks good (improving the xml writing may be done later)

hverhelst:
I submitted a new issue #754 to remind us for the improvements that we can make!

PR #756, opened 2024-12-12
Add cast operators to DScalar1 and DScalar2 types

ArashPartow:
This should cleanly resolve the compilation diagnostic seen due to the lack of conversion operator on the DScalar1/2 types.

PR also reverts the previous change to ExprTk from the following commit: https://github.com/gismo/gismo/commit/64b6edaea7c946a13972c71a5aefd10d1bf2a889


filiatra (approved):
Thank you! We can merge this.

PR #762, opened 2024-12-23
Pygismo docs

filiatra:


hverhelst (approved):
Looks good!
This branch seems to have started from https://github.com/gismo/gismo/pull/723 instead of stable. Maybe we can merge https://github.com/gismo/gismo/pull/723 first.

PR #797, opened 2025-03-12
Bug in interface assembly, after the changes of #779

ckarampa:
**FIXED**: This PR fixes a bug that occurred when using `gsExprAssembler::assembleIfc()`. It was due to some unnecessary indexing happening within `gsExprAssembler::_computePatternIfc()`.

**NEW**: A new constructor of `gsCompositeDomain` is introduced, which accepts a `gsMultiPatch`. Also the corresponding function `gsMultiPatch::domain()` is added. This can be convenient when for example setting the
integration domain of a `gsExprAssembler`, one does not need to create a "dummy" `gsMultiBasis` object just for that.

request:
-----


filiatra (approved):
Ευχαριστούμε !

PR #801, opened 2025-03-20
Fix gsExprAssembler::computePattern

ckarampa:
This PR provides a fix/enhancement in the computation of sparsity patterns when using some version of `gsExprAssembler::assemble / assembleBdr / assembleIfc`.

**FIXED**: When trying to assemble a vector expression on an contact interface ( i.e. using `gsCPPInterface` to handle closest point projections) `computePatternIfc` was being called before the expressions were parsed and this caused a 
bug because the geometry map was not always set, and thus the `gsCPPInterface` could not function properly.

**IMPROVED**:  Since `computePattern` was called regardless of whether a vector or a matrix was beiing assembled, unnecessary loops where performed. This now is avoided.

request:
-----

### DIFF
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -920,6 +920,15 @@ void gsExprAssembler<T>::_computePattern(const expr &... args)
 #pragma omp parallel
 {
     auto arg_tpl = std::make_tuple(args...);
+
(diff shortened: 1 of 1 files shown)


hverhelst (commented):
To me it seems reasonable. 
I leave the approval to @filiatra , since he is the main designer of this code.

filiatra (approved):

PR #802, opened 2025-03-20
Exports class symbol, to be usable in dynamic libraries.

andre-caldas:
FIXED: gcc "undefined reference" errors.

I was getting:
```
undefined reference to `gismo::gsNurbsBasis<double>::knots()'
```
This fixes the problem.

I do not know if there are other classes that need this same fix.

### DIFF
--- a/src/gsNurbs/gsNurbsBasis.h
+++ b/src/gsNurbs/gsNurbsBasis.h
@@ -36,7 +36,7 @@ namespace gismo
 */
 
 template<class T>
-class gsNurbsBasis : public gsRationalBasis<gsBSplineBasis<T> >
+class GISMO_EXPORT gsNurbsBasis : public gsRationalBasis<gsBSplineBasis<T> >
 {
 public:
 


filiatra:
Thnx this is probably a problem only with the gsNurbsBasis class.
There is a warning related to that.
I will check more and get back to this PR.

The warning on GCC is:
```
In file included from gismo/stable/src/gsNurbs/gsNurbsBasis_.cpp:5:
gismo/stable/src/gsNurbs/gsNurbsBasis.h:209:23: warning: type attributes ignored after type is already defined [-Wattributes]
  209 | EXTERN_CLASS_TEMPLATE gsNurbsBasis<real_t>;
      |                       ^~~~~~~~~~~~~~~~~~~~
```

andre-caldas:
> There is a warning related to that.

Very true. I had forgotten about the warning.
GS+mo has a seemingly very clever (I don't really thought about it in deep) way of avoiding the ".hpp" files when linking to the dynamic library. As far as I remember, the warning is caused because a class instatiation that appears after `class gsNurbsBasis` uses `CLASS_TEMPLATE_INST` (and therefore `GISMO_EXPORT`), giving the attribute `(visibility("default"))` to the instantiation without the same attribute being present in the class definition.

andre-caldas:
> As far as I remember, the warning is caused because a class instatiation that appears after `class gsNurbsBasis` uses `CLASS_TEMPLATE_INST` (and therefore `GISMO_EXPORT`), giving the attribute `(visibility("default"))` to the instantiation without the same attribute being present in the class definition.

I suppose this trick of undefining `EXTERN_CLASS_TEMPLATE` should be re-thought. Maybe all those macros should be defined in the same place where `GISMO_BUILD_LIB` is. With the same logic, but no hacks.

filiatra:
This is fixed in PR #811, so closing.

PR #809, opened 2025-04-03
Update CMakeLists.txt

mmoelle1:
CMake 4.x removes support for CMake <3.5


filiatra (approved):
To fix this we use the new syntax of CMake to define a range:

```
if(CMAKE_VERSION VERSION_LESS "3.19")
  cmake_minimum_required(VERSION 2.8.12)
else()
  cmake_minimum_required(VERSION 3.1...3.10)
endif()
```

PR #820, opened 2025-06-07
Introduce C++11 range-based for loops for domain (element) iterators. Fully parallel will openmp

filiatra:


src/gsDomain/gsDomain.h
```diff
@@ -163,6 +176,41 @@ class gsDomain
...
+            iterator domIt = iterator(new gsDomainIteratorEnd<T>(0));
+            return ElementRange( domIt, domIt );
+        }
+#else
+        return ElementRange( beginAll(), endAll() );
+#endif
+    }
+
+    // TO DO:
+    //GISMO_ELEMENT_RANGE_LOOP(allElements, numElements, beginAll, endAll)
```
hverhelst, line 210: Are those a WIP for this PR?

src/gsAssembler/gsExprAssembler.h
```diff
@@ -1586,13 +1564,7 @@ void gsExprAssembler<T>::quPointsWeights(std::vector<gsMatrix<T> >&  cPoints, st
...
 
         // Start iteration over elements of patchInd
-        // use parallel for instead
-#       ifdef _OPENMP
-        domIt += tid;
-        for ( ; domIt<domItEnd; domIt+=nt )
-#       else
-        for (; domIt<domItEnd; ++domIt )
-#       endif
+        for (; domIt<domItEnd; ++domIt ) //todo: parallelize
```
hverhelst: Why don't we use the `for ( auto & elem : m_exprdata->domain().allElements() )` here?

hverhelst (approved):

src/gsDomain/gsDomain.h
filiatra: It is a n idea to do these macros, but on the other hand for boundary elements etc we should probably parallelize accros all patches, which needs more thinking.
So it is not for this PR.

src/gsAssembler/gsExprAssembler.h
filiatra: yes we can, but the function is not correctly parallelized right now (the "count" assumes 1 process).

PR #830, opened 2025-07-21
Fix bug in _expr.npart()

ckarampa:
Fixes a minor bug in the negative part expression's return type. 

### FIX:
* Updated the return type of the `npart` method to use `_expr<Scalar, true>` instead of double, for type consistency.

### IMPROVED:
* Corrected the documentation for the `sqr` method "square" (component-wise) of the expression instead of "square root".

request:
-----


filiatra (changes requested):
Thnx, adding a small unittest will ensure that this will keep working OK in the future.

filiatra (approved):

PR #831, opened 2025-08-12
Add merge function for `gsHTensorBasis`

hverhelst:
Adds a merge function for `gsHTensorBasis`:
```c++
gsHTensorBasis<2> hbasis1(tbasis1);
gsHTensorBasis<2> hbasis2(tbasis2);
hbasis1.merge(hbasis2);
```
This function allows to generate a "union" tree between two `gsHTensorBasis` objects.
Consequently, the function can be used to generate an integration domain for fields represented on two different `gsHTensorBasis` objects.

Test is still missing, but in general we lack tests for `gsHTensorBasis`. Do you have any suggestion, @filiatra @MatuTia @osmihran @dominik-mokris-mtu @AndreaBressan ?

----

NEW:
Merge function for `gsHTensorBasis`

IMPROVED:

FIXED: 

API:

-----

request:
-----


AndreaBressan:
Hi Hugo,

I think the function is fine, but merge for me sounds too much like set union whereas the function transforms the object to the coarsest common refinement of the two bases. What do you think of something along the lines

refine_at_least_as
make_as_fine_as

?

Best and have a nice vacation

Andrea

MatuTia:
Hi Hugo,

I have few questions regarding the function.

- Are we interested in the real types of bases involved? I mean, should the function work if the real types of bases are different? 

In case bases share the real type
- Should merging a base with a copy of itself return the base itself?
- Should `this.merge(other) == other.merge(this)`? In my opinion a function called `merge` should have this property.

However, I have some doubts about what the purpose of the function is, and thus I am out of context.

Best,

Mattia

src/gsDomain/gsHTree.hpp
```diff
@@ -220,6 +220,83 @@ gsHTree<d, Z>::insertBox ( point const & k1, point const & k2,
...
+
+    while (!stack.empty())
+    {
+        std::pair<node*, const node*> current = stack.top();
+        stack.pop();
+
+        node* thisNode = current.first;
+        const node* otherNode = current.second;
+
+        // Case 1: other is a split node, but this is a leaf
```
dominik-mokris-mtu: In G+Smo, we number from 0. (-:

dominik-mokris-mtu:
Hello everyone,
what I'm missing is the explanation "why you'd like us to include them".

When we understand the usecase, we should be in a better position to judge the function (even though it looks already quite good as it is).

hverhelst:
@AndreaBressan @MatuTia @dominik-mokris-mtu , thanks for your comments!
I revised the implementation (simplified A LOT).
There is also a test, an example and the PR description motivates why we want this (@dominik-mokris-mtu ).
The function also has a `static` version (allowing `merge(basis1,basis2)`) and the property `basis1.merge(basis2)==basis2.merge(basis2)` is tested in the test `MergeIsCommutative`. I assume the real types are the same for both bases (@MatuTia ).
The name `merge` is kept for now, @AndreaBressan , but still not decided.

CC @filiatra

filiatra (approved):

PR #837, opened 2025-10-09
Feature/multipatch single file

clebreto:
# Unstructured Grid VTK Export - Summary

## Overview
Successfully implemented single-file VTK unstructured grid export for multi-patch geometries in G+Smo, with support for both ASCII and binary (base64) formats.

### Functions Added
1. **`gsWriteParaviewUnstructuredGrid(const gsMultiPatch<T>&, string, unsigned, bool)`**
   - Exports multi-patch geometry to single .vtu file
   - Includes PatchID cell data for patch identification
   
2. **`gsWriteParaviewUnstructuredGrid(const gsField<T>&, string, unsigned, bool)`**
   - Exports field (geometry + solution) to single .vtu file
   - Includes both geometry and field data

### Features
- **Format Support**: Both ASCII (default) and base64 binary encoding
- **Backward Compatibility**: Original multi-file structured grid export still available
- **Integration**: Works with existing `gsView` tool via `--base64` flag
- **Pixi Support**: Project configured with pixi for reproducible builds

#### ASCII Format
- **Command**: `pixi run gsview-ascii`
- **Output File**: `car_ascii.vtu` (129M)
- **Points**: 1,518,800
- **Cells**: 1,230,228
- **Format**: `format="ascii"` in DataArray elements

#### Binary Format
- **Command**: `pixi run gsview-binary`
- **Output File**: `car_binary.vtu` (87M)
- **Points**: 1,518,800
- **Cells**: 1,230,228
- **Format**: `format="binary"` in DataArray elements (base64 encoded)
- **File Size Reduction**: 32% smaller than ASCII (87M vs 129M)

### Format Comparison

| Feature | ASCII | Binary (Base64) |
|---------|-------|----------------|
| File Size (car.xml) | 129 MB | 87 MB |
| Compression | None | 32% reduction |
| Human Readable | Yes | No |
| ParaView Compatible | Yes | Yes |
| Performance | Slower I/O | Faster I/O |

### Build with Pixi
```bash
pixi install
pixi run configure
pixi run build
```

#### ASCII Format (default)
```bash
pixi run gsview-ascii
# or
./build/bin/gsView -i filedata/geo/car.xml -o car_ascii
```

#### Binary Format
```bash
pixi run gsview-binary
# or
./build/bin/gsView -i filedata/geo/car.xml -o car_binary --base64
```

#### Export Geometry Only
```cpp
#include <gsIO/gsWriteParaview.h>

// Load multi-patch geometry
gsMultiPatch<> mp;
gsReadFile<>("geometry.xml", mp);

// Export to ASCII (default)
gsWriteParaviewUnstructuredGrid(mp, "output", 1000);

// Export to binary
gsWriteParaviewUnstructuredGrid(mp, "output", 1000, true);
```

#### Export with Field
```cpp
#include <gsIO/gsWriteParaview.h>

// Create field with geometry and solution
gsField<> field(patches, solution);

// Export to ASCII
gsWriteParaviewUnstructuredGrid(field, "output", 1000);

// Export to binary
gsWriteParaviewUnstructuredGrid(field, "output", 1000, true);
```

### VTU File Format
- **Header**: Standard VTK XML header
- **UnstructuredGrid**: Single piece containing all patches
- **Points**: All control points from all patches
- **Cells**: All elements (quads or hexahedra) from all patches
- **Cell Data**: 
  - PatchID: Integer identifying which patch each cell belongs to
  - Field data (if applicable)

### PVD File Format
```xml
<?xml version="1.0"?>
<VTKFile type="Collection" version="0.1">
<Collection>
<DataSet file="output.vtu"/>
</Collection>
</VTKFile>
```

## Key Benefits

1. **Single File**: Easier to manage than multiple files per patch
2. **ParaView Ready**: Direct visualization without collection files
3. **Patch Tracking**: PatchID cell data preserves patch information
4. **Flexible**: Supports both ASCII and binary formats
5. **Efficient**: Binary format reduces file size by ~32%
6. **Compatible**: Works with existing G+Smo workflows

## Performance Notes

- **Binary format recommended** for large models (>1M points)
- **ASCII format useful** for debugging and small models
- Base64 encoding provides good balance of compression and compatibility
- Single file reduces filesystem overhead compared to multi-file approach

### Core Implementation
- `src/gsIO/gsWriteParaview.h` - Function declarations
- `src/gsIO/gsWriteParaview.hpp` - Implementation
- `src/gsIO/gsWriteParaview_.cpp` - Template instantiations

### Tools
- `examples/gsView.cpp` - Already integrated (uses new functions)

### Configuration
- `pixi.toml` - Project configuration with build and export tasks

## Next Steps

The implementation is complete and tested. You can now:
1. Open `car_ascii.pvd` or `car_binary.pvd` in ParaView
2. Visualize the 15,188-patch car model as a single unstructured grid
3. Use the PatchID cell data to color/filter by patch
4. Compare file sizes and load times between ASCII and binary formats
5. Integrate the functions into your own G+Smo applications

## Conclusion

The single-file unstructured grid export is production-ready and provides a significant improvement over the previous multi-file structured grid approach. The binary format option offers substantial file size savings while maintaining full ParaView compatibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,9 @@ CMakeFiles/
 
 #Symbolic link Eigen library
 external/Eigen
+
+# vscode
+.cache
+
+# usual cmake build directory
+build/
--- a/examples/heatEquation_example.cpp
+++ b/examples/heatEquation_example.cpp
@@ -22,9 +22,11 @@ int main(int argc, char *argv[])
 {
     bool plot = false;
     bool export_base64 = false;
+    int numSteps = 40;
     gsCmdLine cmd("Testing the heat equation.");
     cmd.addSwitch("plot", "Plot the result in ParaView.", plot);
     cmd.addSwitch("base64", "write in binary format", export_base64);
+    cmd.addInt( "n", "numsteps", "Number of time steps.", numSteps);
     try { cmd.getValues(argc,argv); } catch (int rv) { return rv; }
 
     // Source function
@@ -93,7 +95,6 @@ int main(int argc, char *argv[])
     gsMatrix<> Sol, Rhs;
     int ndof = assembler.numDofs();
     real_t endTime = 0.1;
-    int numSteps = 40;
     Sol.setZero(ndof, 1); // Initial solution
 
     real_t Dt = endTime / numSteps ;
--- a/examples/heatEquation_example.cpp
+++ b/examples/heatEquation_example.cpp
@@ -11,6 +11,7 @@
     Author(s): S. Moore, A. Mantzaflaris
 */
 
+#include "gsCore/gsDebug.h"
 #include <gismo.h>
 
 
@@ -20,8 +21,10 @@ using namespace gismo;
 int main(int argc, char *argv[])
 {
     bool plot = false;
+    bool export_base64 = false;
     gsCmdLine cmd("Testing the heat equation.");
     cmd.addSwitch("plot", "Plot the result in ParaView.", plot);
+    cmd.addSwitch("base64", "write in binary format", export_base64);
     try { cmd.getValues(argc,argv); } catch (int rv) { return rv; }
 
     // Source function
@@ -99,6 +102,7 @@ int main(int argc, char *argv[])
     gsParaviewCollection collection(baseName);
     collection.options().setInt("numPoints", 1000);
     collection.options().setInt("precision", 5);
+    collection.options().setSwitch("base64", export_base64);
 
     if ( plot )
     {
--- a/src/gsIO/gsParaviewCollection.cpp
+++ b/src/gsIO/gsParaviewCollection.cpp
@@ -11,6 +11,7 @@
     Author(s): A. Mantzaflaris, C. Karampatzakis
 */
 
+#include "gsCore/gsDebug.h"
 #include <gsIO/gsParaviewCollection.h>
 
 
@@ -86,5 +87,13 @@ namespace gismo
        
         m_dataset = gsParaviewDataSet(name, geometry, m_evaluator, m_options);
     }
-}
-
+    void gsParaviewCollection::writeToDisk() const {
+        gsInfo << "Exporting to " << m_filename << "\n";
+      std::ofstream f(m_filename.c_str());
+      GISMO_ASSERT(f.is_open(), "Error creating " << m_filename);
+      f << mfile.str();
+      f << "</Collection>\n";
+      f << "</VTKFile>\n";
+      f.close();
+    }
+} // namespace gismo
--- a/src/gsIO/gsWriteParaview.h
+++ b/src/gsIO/gsWriteParaview.h
@@ -158,7 +158,7 @@ void gsWriteParaviewBezier(const gsMultiPatch<T> & mPatch, std::string const & f
 /// \param fn filename where paraview file is written (without extension)
 /// \param npts number of points used for sampling each patch
 template<class T>
-void gsWriteParaviewUnstructuredGrid(const gsMultiPatch<T> & mPatch, std::string const & fn, unsigned npts = NS);
+void gsWriteParaviewUnstructuredGrid(const gsMultiPatch<T> & mPatch, std::string const & fn, unsigned npts = NS, bool export_base64 = false);
 
 /// \brief Export a field (solution on geometry) to a single unstructured grid file (all patches in one .vtu file)
 ///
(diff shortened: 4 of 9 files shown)


ckarampa:
Hello and thanks for the contribution! What you have done looks really interesting. I would like to point out, that now that you have implemented the unstructured grid export to a single file, it would make sense for us to sample each element, and not each patch. Especially with THB - Splines, the elements of one patch might have vastly different size, and it would take a very fine (and redundant) sampling to capture the solution in the small ones. On the other hand, if we sample each element, the sampling automatically adjusts to the mesh and should improve performance and reduce file sizes for THB while not affecting uniform meshes.

hverhelst:
> Hello and thanks for the contribution! What you have done looks really interesting. I would like to point out, that now that you have implemented the unstructured grid export to a single file, it would make sense for us to sample each element, and not each patch. Especially with THB - Splines, the elements of one patch might have vastly different size, and it would take a very fine (and redundant) sampling to capture the solution in the small ones. On the other hand, if we sample each element, the sampling automatically adjusts to the mesh and should improve performance and reduce file sizes for THB while not affecting uniform meshes.

I do really like the new structure, reducing multi-patch output to a single file, but I support @ckarampa 's comment that sampled elements (or integration points) should be the standard instead of patches.

.gitignore
```diff
@@ -42,3 +42,9 @@ CMakeFiles/
 
 #Symbolic link Eigen library
 external/Eigen
+
+# vscode
+.cache
+
+# usual cmake build directory
+build/
```
hverhelst, line 50: Could we also use `build*`?
This would capture `build_deb`, `build_rel`, what people often use.

examples/heatEquation_example.cpp
```diff
@@ -11,6 +11,7 @@
     Author(s): S. Moore, A. Mantzaflaris
 */
 
+#include "gsCore/gsDebug.h"
```
hverhelst, line 14: Is this needed?

src/gsIO/gsParaviewCollection.cpp
```diff
@@ -11,6 +11,7 @@
     Author(s): A. Mantzaflaris, C. Karampatzakis
 */
 
+#include "gsCore/gsDebug.h"
```
hverhelst, line 14: Is this needed?

src/gsIO/gsWriteParaview.h
hverhelst, line 1: How difficult is it to export the control net or elements in a similar fashion? 
The existing `gsWriteParaview` for `gsMultiPatch` can optionally plot those, though also exported to different files.

clebreto:
Hello !

Thanks for this super quick review.

I should have made it a draft, we will discuss it at length with @filiatra !

src/gsIO/gsWriteParaview.h
clebreto: We will look into doing that.

ckarampa: Not that any one asked (:-P) , but I think that the control grid / element mesh should indeed be in separate files, so that they will end up in different Blocks in paraview, and their visibility can be toggled individually

hverhelst:
> Hello !
> 
> Thanks for this super quick review.
> 
> I should have made it a draft, we will discuss it at length with @filiatra !

No worries, thanks for opening it anyway!

src/gsIO/gsWriteParaview.h
clebreto: I feel that having each patch control points in its own file would go against the goal we are trying to achieve here: providing Paraview with a single file to improve the scene rendering speed.

Filling a folder with hundreds to the ten thousands of files does not seem very effective.

Paraview provides vtkMultiBlockDataSet.

I would rather look into this direction.

ckarampa: Yes I agree with you, maybe my phrasing wasn't clear. I meant that there should be one file for geometry, one file for the control grid and one for the element mesh ( not one file per element etc. ). My bad

clebreto: Then great ! (I read that vtkMultiBlockDataSet might not be super efficient for large amounts of patches, let's check.)

clebreto: The vtkMultiBlockDataset is indeed very slow to load with 15000 patches. It is not straightforward to hide and show the patches. I do not think it is the way to go. I can put the code in a branch if you feel like testing.

hverhelst:
This PR is included in #876 , so will be closed when #876 merges.

filiatra:
OK, the code is now committed in dev in a bigger PR -- closing the PR.

PR #838, opened 2025-10-21
New implementation of admissibility algorithms

hverhelst:
Adds a new implementation of adaptive meshing and element representation, using the classes `gsElement`, `gsHElement`, `gsElementHelper`, `gsHElementHelper` and `gsHElementMarker`, instead of the old implementation on `gsHBox`, `gsHBoxContainer` and `gsAdaptiveMeshing`.

## Why changing the old implementation
The old implementation used `std::vector` as a container for boxes. These turned out to contain a lot of duplicates. As a result, iteration over a support extension would result in many duplicate operations. 

## What is the difference between `gsElement` and `gsHElement`?
The `gsElement` class is a smart representation of elements. It takes two box coordinates and implements comparison operators. The `gsElementHelper` implements the support extension (requires a basis) and an explode function. These are almost some example operations for a normal element, and can be extended in the future.
The `gsHElement` class inherits from `gsElement` and adds a level. Since there is a level, all operations needed for admissible refinement (according to [1]) can be implemented.

[1] Bracco, C., Giannelli, C., & Vázquez, R. (2018). Refinement algorithms for adaptive isogeometric methods with hierarchical splines. axioms, 7(3), 43.

-------
NEW: 
New element representations for hierarchical meshes: `gsElement`, `gsHElement`, `gsElementHelper`, `gsHElementHelper` and `gsHElementMarker`. 

IMPROVED:
Paraview plot of boxes and elements

-------
-----


hverhelst:
**Question**: @filiatra @MatuTia, as these new classes replace `gsHBox` etc., we can deprecate the old classes `gsHBox`, `gsHBoxContainer` and `gsAdaptiveMeshing` and remove them in a future version. What do you think?

filiatra (dismissed):

filiatra:
> **Question**: @filiatra @MatuTia, as these new classes replace `gsHBox` etc., we can deprecate the old classes `gsHBox`, `gsHBoxContainer` and `gsAdaptiveMeshing` and remove them in a future version. What do you think?

This would be great !
Maybe just a next PR removing the gsHBox etc

filiatra (dismissed):
Something breaks:
https://cdash-ci.irisa.fr/viewBuildError.php?buildid=160099

filiatra (approved):

PR #840, opened 2025-10-22
Add #include<cstdint> explicitly

MatuTia:
This PR explicitly adds `#include <cstdint>` for GCC 15 compatibility.
 
NEW: Explicit `#include <cstdint>` in `gsForwardDeclarations`; no longer included transitively by standard C++ headers.


filiatra (approved):
OK thank you!

PR #856, opened 2026-02-03
fix output in basis example

Drualis:
To know if the basis function 2 is active, one has the query that. There is no need to dump the variable "active" a second time.


stefantakacs (approved):
LGTM

PR #859, opened 2026-02-17
Add gsSubdivisionScheme and two example implementors

Linus-Mussmaecher:
NEW: An abstract class `gsSubdivisionScheme` that provides a common interface for different subdivision schemes.
NEW: A concrete class `gsCatmullClark` that implements `gsSubdivisionScheme` and contains the Catmull-Clark-Subdivision.
NEW: A concrete class `gsDooSabin` that implements `gsSubdivisionScheme` and contains the Doo-Sabin-Subdivision.
NEW: An example `subdivisionSurfaces_example` that shows how to use these new classes.

The code in the header class is extensively documented via Doxygen.
The code in the implementors is directly taken from @filiatra 's `gsSubdivScheme`.

TODO: Review & test the copied code.
TODO: Implement the `check_mesh` function for the subdivision schemes (optional).
TODO: Move the other subdivision schemes from `gsSubdivScheme` over to new classes implementing `gsSubdivisionScheme`.
TODO: Delete the old class `gsSubdivScheme` and its example `subdivSurfaces_example`.


examples/subdivisionSurfaces_example.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // TODO: Fit other schemes into this class hierarchy and implement the example properly.
+    gsSubdivisionScheme* scheme;
+
+    if (scheme_name == "Catmull-Clark"){
+            scheme = new gsCatmullClark();
+    } else if (scheme_name == "Doo-Sabin"){
+            scheme = new gsDooSabin();
+    } else {
+            scheme = new gsCatmullClark();
+    }
```
hverhelst: Another formatting thing (which is not changed automatically) is that we usually (not strictly) open and close curly brackets on a new line

PR #864, opened 2026-03-04
Parasolid closest point computation for parameter correction

dominik-mokris-mtu:
approval) from @gismo/admins.

Code submitted to the dev branch should be clean, well-documented
and free of bugs.

NEW: Function `gsClosestParam` to compute the parameter values of the closest point on a tensor-product surface using Parasolid. This is a debugged part from the stale branch [Parasolid-parameter-correction](https://github.com/gismo/gismo/tree/Parasolid-parameter-correction).

IMPROVED: Several utility functions have been moved from _gsWriteParasolid_ to newly-created _gsPKUtils_.

For each new/improved/fixed/api change use a new line with the
respective word prefix in capital.

request:
Kind of: they are in a private repository.
-----

### DIFF
--- a/optional/gsParasolid/gsClosestPoint.cpp
+++ b/optional/gsParasolid/gsClosestPoint.cpp
@@ -1,0 +1,37 @@
+/** @file gsClosestPoint.cpp
+
+    @brief Provides instantization of gsClosestPoint functions.
+
+    This file is part of the G+Smo library. 
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+    
+    Author(s): D. Mokris
+*/
+
+#include <gsCore/gsTemplateTools.h>
+
+#include <gsParasolid/gsClosestPoint.h>
+#include <gsParasolid/gsClosestPoint.hpp>
+
+namespace gismo
+{
+namespace extensions
+{
+TEMPLATE_INST PK_VECTOR_t
+gsPK_VECTOR(const gsVector<real_t, 3>& gsVector);
+
+TEMPLATE_INST bool
+gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
+               const gsVector<real_t, 3>& gsPoint,
+               gsVector<real_t, 2>& gsResult);
+
+TEMPLATE_INST bool
+gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
+               const gsMatrix<real_t>& gsPoints,
+               gsMatrix<real_t>& gsResults);
+
+} // namespace extensions
+} // namespace gismo
--- a/optional/gsParasolid/gsClosestPoint.cpp
+++ b/optional/gsParasolid/gsClosestPoint.cpp
@@ -11,8 +11,6 @@
     Author(s): D. Mokris
 */
 
-#include <gsCore/gsTemplateTools.h>
-
 #include <gsParasolid/gsClosestPoint.h>
 #include <gsParasolid/gsClosestPoint.hpp>
 
--- a/optional/gsParasolid/gsClosestPoint.cpp
+++ b/optional/gsParasolid/gsClosestPoint.cpp
@@ -26,12 +26,14 @@ gsPK_VECTOR(const gsVector<real_t, 3>& gsVector);
 TEMPLATE_INST bool
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsVector<real_t, 3>& gsPoint,
-               gsVector<real_t, 2>& gsResult);
+               gsVector<real_t, 2>& gsResult,
+               bool performance = true);
 
 TEMPLATE_INST bool
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsMatrix<real_t>& gsPoints,
-               gsMatrix<real_t>& gsResults);
+               gsMatrix<real_t>& gsResults,
+               bool performance = true);
 
 } // namespace extensions
 } // namespace gismo
--- a/optional/gsParasolid/gsClosestPoint.cpp
+++ b/optional/gsParasolid/gsClosestPoint.cpp
@@ -21,13 +21,13 @@ namespace extensions
 TEMPLATE_INST PK_VECTOR_t
 gsPK_VECTOR(const gsVector<real_t, 3>& gsVector);
 
-TEMPLATE_INST bool
+TEMPLATE_INST PK_ERROR_code_t
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsVector<real_t, 3>& gsPoint,
                gsVector<real_t, 2>& gsResult,
                bool performance);
 
-TEMPLATE_INST bool
+TEMPLATE_INST PK_ERROR_code_t
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsMatrix<real_t>& gsPoints,
                gsMatrix<real_t>& gsResults,
--- a/optional/gsParasolid/gsClosestPoint.cpp
+++ b/optional/gsParasolid/gsClosestPoint.cpp
@@ -25,13 +25,13 @@ TEMPLATE_INST PK_ERROR_code_t
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsVector<real_t, 3>& gsPoint,
                gsVector<real_t, 2>& gsResult,
-               range::opt optimise);
+               range_opt optimise);
 
 TEMPLATE_INST PK_ERROR_code_t
 gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
                const gsMatrix<real_t>& gsPoints,
                gsMatrix<real_t>& gsResults,
-               range::opt optimise);
+               range_opt optimise);
 
 } // namespace extensions
 } // namespace gismo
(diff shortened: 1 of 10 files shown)


optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+           error status.
+        */
+        template <class T>
+        bool gsClosestParam(const gsTensorBSpline<2, T>& gsBSurf,
+                            const gsVector<T, 3>& gsPoint,
+                            gsVector<T, 2>& gsResult,
+                            bool performance = true);
+
+        /**
+           For each three-dimensional point represented as a row of \a
```
scholzfelix, line 57: The documentation strings of the two gsClosestParam functions seem to be not up to date.

In this one it describes multiple rows in gsPoints, but this function only takes a single point as a gsVector.

Also `gsResult` instead of `gsResults`

optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+                            const gsMatrix<T>& gsPoints,
+                            gsMatrix<T>& gsResults,
+                            bool performance = true);
+
+        /**
+           Finds the point on \a gsBSurf that is closest to \a point
+           and writes it into \a result.
+           Returns Parasolid error status.
+
+           TODO: Not implemented, yet.
```
scholzfelix: Here it says not implemented, but you implemented it :)

optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+
+        /**
+           Finds the point on \a gsBSurf that is closest to \a point
+           and writes its parameter values into \a result. The
+           optional parameter \a performance decides between
+           performance (true) and precision (false). Returns Parasolid
+           error status.
+        */
+        template <class T>
+        bool gsClosestParam(const gsTensorBSpline<2, T>& gsBSurf,
```
scholzfelix: I think the return type of `gsClosestParam` should be `PK_ERROR_code_t`

optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+                            const gsVector<T, 3>& gsPoint,
+                            gsVector<T, 2>& gsResult,
+                            bool performance = true);
```
scholzfelix: It would be more elegant to have the possible values of `performance` as an `enum`, similar to how it is done in `gsAssemblerOptions.h`. The we could have something like `optimise::performance` and `optimise::accuracy`

optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+
+        /**
+           Finds the point on \a gsBSurf that is closest to \a point
+           and writes it into \a result.
+           Returns Parasolid error status.
+
+           TODO: Not implemented, yet.
+        */
+        template <class T>
+        bool gsClosestPoint(const gsTensorBSpline<2, T>& gsBSurf,
```
scholzfelix: If the function is not implemented, do we need it?

optional/gsParasolid/gsClosestPoint.cpp
```diff
@@ -0,0 +1,39 @@
...
+    This file is part of the G+Smo library. 
+
+    This Source Code Form is subject to the terms of the Mozilla Public
+    License, v. 2.0. If a copy of the MPL was not distributed with this
+    file, You can obtain one at http://mozilla.org/MPL/2.0/.
+    
+    Author(s): D. Mokris
+*/
+
+#include <gsCore/gsTemplateTools.h>
```
scholzfelix: This include seems to be unused

optional/gsParasolid/gsClosestPoint.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace extensions
+{
+TEMPLATE_INST PK_VECTOR_t
+gsPK_VECTOR(const gsVector<real_t, 3>& gsVector);
+
+TEMPLATE_INST bool
+gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
+               const gsVector<real_t, 3>& gsPoint,
+               gsVector<real_t, 2>& gsResult,
+               bool performance = true);
```
scholzfelix: The default value should not be set here as it is already set in `gsClosestPoint.h`.

optional/gsParasolid/gsClosestPoint.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+TEMPLATE_INST bool
+gsClosestParam(const gsTensorBSpline<2, real_t>& gsBSurf,
+               const gsMatrix<real_t>& gsPoints,
+               gsMatrix<real_t>& gsResults,
+               bool performance = true);
```
scholzfelix: Same as above

optional/gsParasolid/gsClosestPoint.h
```diff
@@ -0,0 +1,81 @@
...
+struct PK_VECTOR_s;
+typedef struct PK_VECTOR_s PK_VECTOR_t;
+
+namespace gismo
+{
+    namespace extensions
+    {
+        /// Converts a G+Smo vector to a Parasolid vector.
+        template <class T>
+        PK_VECTOR_t gsPK_VECTOR(const gsVector<T, 3>& gsVector);
```
scholzfelix, line 35: At some point it would be nice to move all the utility functions to a separate file, also the ones defined in `gsWriteParasolid.h`. 
I was at first confused, why you needed to include `gsWriteParasolid.h` in `gsClosestPoint.hpp`.

scholzfelix (changes requested):

optional/gsParasolid/gsClosestPoint.cpp
dominik-mokris-mtu: Good point, solved in b5515eb93e552e90d7c19978dbe061b8f702cc9e.

optional/gsParasolid/gsClosestPoint.cpp
dominik-mokris-mtu: Same as above. (-:

optional/gsParasolid/gsClosestPoint.cpp
dominik-mokris-mtu: You are right, this was copy-pasting on my side. Solved in 0b2bd85e072778b7ad1c902ecec81e4912181ed3.

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: Done a313b1810cdc9d917031cff19b8438ae3f2d0e4c . The way I did it seems like over-engineering, though. What do you think?

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: Done in 8e69fa30a95283ce3b21d9f7b5bdecb1f1ef69dd .

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: Removed in e90ca1dd1209b946522270910f40b4d0e1f28d38 .

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: Not really. But it makes you wonder why the file is called that way. I'll keep the name in case we decide to implement the function later on.

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: I have updated the documentation but I think it was more or less correct: there is a version working with just one point (using `gsVector` and names in singular) and another one working with several points in parallel (which uses `gsMatrix` and names in plural. Or am I missing something?

optional/gsParasolid/gsClosestPoint.h
scholzfelix: I like it! I just added a tiny improvement suggestion as a comment to the code [here](https://github.com/gismo/gismo/commit/a313b1810cdc9d917031cff19b8438ae3f2d0e4c#r179472832)

If you want to have less overhead you can also remove the `struct` and only use the `enum`.

optional/gsParasolid/gsClosestPoint.h
scholzfelix: Sorry, I think was confused about something here. Probably related to docstrings in Lisp being below the function signature.
You are right that it is correct!

optional/gsParasolid/gsClosestPoint.h
scholzfelix: Not sure if linking the comment worked here. 
What I mean by tiny improvement is that I think in line 51 inline 92 you can directly write
`options.opt_level = optimise;`, because you already chose the values in the enum do be the correct Parasolid constants.

scholzfelix (approved):

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: I started this in 76d8a6ac43e18a452489ef616559e8f8c2736b8e . It took some effort to make it compile and the order of includes might still make trouble on other computers. Let's see how it develops.

optional/gsParasolid/gsClosestPoint.h
dominik-mokris-mtu: Oh, quite clever!

dominik-mokris-mtu: > I like it! I just added a tiny improvement suggestion as a comment to the code [here](https://github.com/gismo/gismo/commit/a313b1810cdc9d917031cff19b8438ae3f2d0e4c#r179472832)
> 
> If you want to have less overhead you can also remove the `struct` and only use the `enum`.

Done in 226ac406cb0f0aaab4fdec94dea0c412d0ca5892 , although the name of the `enum` can probably be further optimised.

dominik-mokris-mtu:
@filiatra : Felix seems to be happy with my changes. How about you?

filiatra (approved):

Issue #862, opened 2026-02-25
Fitting with non-clamped knots crashes due to a singular matrix.


dominik-mokris-mtu:
**Describe the bug**
When performing least-squares fitting with a tensor-product B-splines with non-clamped knots, the matrix assembled in `gsFitting<T>::compute` is singular. This leads first to
```
Warning: The preconditioner failed. Aborting.
```
and later to a crash.

**To Reproduce**
Make a cpp-file from the following code, compile it and run it.
```cpp
#include <gismo.h>

using namespace gismo;

int main()
{
	gsFileData<> in("fitting/deepdrawingC.xml");
	gsMatrix<> uv, xyz;
	in.getId<gsMatrix<>>(0, uv);
	in.getId<gsMatrix<>>(1, xyz);

    std::vector<real_t> knots = {-2, -1.75, -1.5, -1.25, -1, -0.75, -0.5, -0.25, 0, 0.25, 0.5, 0.75, 1, 1.25, 1.5, 1.75, 2};
	gsKnotVector<> u_knots(3, knots.begin(), knots.end());
	gsKnotVector<> v_knots(-1, 1, 15, 4);
	gsTensorBSplineBasis<2> basis(u_knots, v_knots);

	gsFitting<> fitting(uv, xyz, basis);
	fitting.compute(1e-6);
	gsInfo << *fitting.result() << std::endl;

	return 0;	
}
```

**Expected behavior**
The fitting should yield a tensor-product surface that is then printed on the command line.

**Hints**

1. When constructing the `u_knots` with `gsKnotVector<> u_knots(-1, 1, 15, 4);` (which is slightly different), the code runs.
2. The problem has appeared with #779 . When using commit df1a96f77ead90b383b4ea3e403e4ce707a8fcfe, which should be the last one before the PR, the code runs.
3. <img width="969" height="727" alt="Image" src="https://github.com/user-attachments/assets/1beeddc9-c8ce-4b1a-93f4-55e880bb5843" />

**Desktop (please complete the following information):**
 - OS: Linux SUSE Enterprise
 - Branch/fork used: dev. See also hint number 2.

dominik-mokris-mtu:
@scholzfelix: This is the issue we've been discussing yesterday.

dominik-mokris-mtu:
@hverhelst: Do you have an idea what could be going wrong here?

hverhelst:
Interesting! I'll have a look

hverhelst:
[...]

PR #865, opened 2026-03-06
Hotfix issue 862: Knot domain iterator fix

hverhelst:
This PR fixes issue #862 .
The problem was that upon a `reset()`, in commit `df1a96f77ead90b383b4ea3e403e4ce707a8fcfe` the knot vector would go back to `domainUBegin()` and in commit `1d7b2b01947ddc2037fc541076a9eba91dc87f16` this changed to `m_upos=0` which is the first unique knot, which in case of non-clamped knot vectors is a ghost knot.

This happened because `nextLexicographic` and `nextLexicographicIter` called, respectively, `cur[i]=start[i]` and `cur[i].reset()` to reset a current dimension. This lexicographic iterator is used for the `gsTensorDomainIterator`, which is constructed via breaks for `gsKnotVector`s (is my guess), which sets for `start` the `domainUBegin` of the knot vectors. For non-clamped knot vectors, `cur[i].reset()` did not yield the same result as `cur[i].domainUBegin()` which it should do.

@dominik-mokris-mtu @filiatra , what do you think about this? 

---

NEW: 
Added unittests for non-clamped knot vector iteration, according to #862 

IMPROVED: 

FIXED: 
non-clamped knot vector iteration

API:

---
request:
-----

### DIFF
--- a/src/gsDomain/gsKnotDomainIterator.h
+++ b/src/gsDomain/gsKnotDomainIterator.h
@@ -28,7 +28,7 @@ private:
     typedef typename gsDomainIterator<T>::uPtr domainIter;
 
     // Data members
-    knotIterator m_it, m_itEnd;
+    knotIterator m_it, m_itBegin, m_itEnd;
 
 public:
 
@@ -36,6 +36,7 @@ public:
     :
     gsDomainIterator<T>(start ? 0 : _knots.numElements()),
     m_it(start ? _knots.domainUBegin() : _knots.domainUEnd()),
+    m_itBegin(_knots.domainUBegin()),
     m_itEnd(_knots.domainUEnd())
     {
 
@@ -71,7 +72,7 @@ public:
     // Documentation in gsDomainIterator.h
     void reset() override
     {
-        m_it.reset();
+        m_it = m_itBegin;
     }
 
     gsVector<T> lowerCorner() const override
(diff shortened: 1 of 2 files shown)


dominik-mokris-mtu (approved):
Thanks for looking into the problem. From my viewpoint it is good. I'm leaving two comments to think about, though.

dominik-mokris-mtu:
Sorry to be that kind of guy: any news here? (-:

hverhelst:
> Sorry to be that kind of guy: any news here? (-:

Hi! I was busy with some deadline, sorry for my late reply.
I planned to look at your comments but they disappeared.
Is it still relevant?

dominik-mokris-mtu:
> > Sorry to be that kind of guy: any news here? (-:
> 
> Hi! I was busy with some deadline, sorry for my late reply. I planned to look at your comments but they disappeared. Is it still relevant?

No worries. I'd suggest merging and not worrying too much about them.

Just in case, I can still see them:
image

filiatra (approved):
I think the PR is OK.
Note that the knot iterator would iterate over ghost knots as well.

We do assuem that the knotvector does not change during the use of the iterator, otherwise everything will crash..

dominik-mokris-mtu:
Thanks, everyone!

PR #871, opened 2026-03-14
Fix missing cstdint header

topazus:
Fix missing cstdint header

- https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes
- https://gcc.gnu.org/gcc-15/porting_to.html#header-dep-changes


filiatra (approved):

filiatra:
Thank you !

PR #875, opened 2026-04-02
Fix a bug in creation of high-dimension control nets

Linus-Mussmaecher:
This pull request fixes the following bug:

FIX: Calling `gsWriteParaview` with `control_net=true` on a `gsMultiPatch` whose `geoDim()` returns 4 (or higher) and that has at least one control point whose 4th coordinate value is non-zero causes an invalid free (causing a SIGABRT) at the end of the write call.

Here is a minimal example:
```cpp
gsMultiPatch<> mp(*gsNurbsCreator<>::BSplineSquareDeg(2));
mp.computeTopology();
mp.embed(4);

for (size_t p = 0; p < mp.nPatches(); ++p)
    mp.patch(p).coefs().col(3).setConstant(1.0);

gsWriteParaview(mp, "cnet_4d", 100, false, true);
```

The cause is that the code in `src/gsIO/gsWriteParaview.hpp`, function `writeSingleControlNet`, checks for `geoDim() > 3` too late, creating a `gsMesh` earlier and triggering an asserting in the constructor of `gsVertex`.
This pull request moves the check to an earlier part of the method to avoid these aborts.

I am reasonably certain this fixes the error while not changing any behaviour vor `geoDim() <= 3`, but would appreciate if someone could look over it to see if the original intention of the code is intact.

### DIFF
--- a/src/gsIO/gsWriteParaview.hpp
+++ b/src/gsIO/gsWriteParaview.hpp
@@ -248,9 +248,20 @@ void writeSingleControlNet(const gsGeometry<T> & Geo,
                            std::string const & fn)
 {
     const int d = Geo.parDim();
+    const unsigned n = Geo.geoDim();
+
+    // If the dimension of the geometry is too high, fall back to writing
+    // the control points as an unstructured point cloud instead.
+    if (n > 3)
+    {
+        gsDebug<<"Writing 4th coordinate\n";
+        const gsMatrix<T> & cp = Geo.coefs();
+        gsWriteParaviewPoints<T>(cp.transpose(), fn );
+        return;
+    }
+
     gsMesh<T> msh;
     Geo.controlNet(msh);
-    const unsigned n = Geo.geoDim();
     if ( n == 1 )
     {
         gsMatrix<T> anch = Geo.basis().anchors();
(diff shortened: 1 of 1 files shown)


Linus-Mussmaecher:
I only tested for 4 dimensions previously, but the AI is correct that >4 is actually not supported at all. I updated the function accordingly and also slightly tweaked the error message sent by `gsWriteParaviewPoints` from stating that only dimension 1-3 are supported to stating that only dimensions 1-4 are supported.

Would still appreciate a second pair of eyes on this.

Small edit: I just now realized that I also ran my autoformatter on the changed file, causing quite a lot of small changes (mostly removing double spaces before `=` it seems). This blows the number of changed lines far beyond the ~10 lines of functional changes.

filiatra (changes requested):
Thnx for the PR !

Please make a new PR that contains your fixes but not any formatting changes.

You could open a separate PR for formatting changes only (these are welcome as well!).
The reason is that the changelogs are poluted by the formating updates and makes it very hard to track the real changes in the code.

Linus-Mussmaecher:
This is understandable, I have created a new pull request in #877 and will delete this branch.

PR #885, opened 2026-06-26
Provide third derivatives evaluation to gsTensorBSplineBasis

dominik-mokris-mtu:
NEW: deriv3, deriv3_into, deriv3Single, deriv3Single_into for gsTensorBSplineBasis

# Suggested change
This code provides evaluation of third derivatives for tensor-product B-spline basis.

# Motivation
We have good experience with using third derivatives for surface smoothing, which we do in a private repository.

#  Why it's not a big deal
The implementation is very analogous to how second derivatives are handled. It is mostly taken from 
the branch [sandra](https://github.com/gismo/gismo/tree/sandra) that we (@Sandra-Merchel-MTU and me) did not merge in 2020.

request:
- [X] Have you added an explanation of what your changes do and why you'd like us to include them? _See above._
- [X] Have you documented any new codes using Doxygen comments? _Analogously to deriv2._
- [X] Have you written new tests or examples for your changes? _Yes, there are three new unittests._
-----

### DIFF
--- a/src/gsCore/gsBasis.h
+++ b/src/gsCore/gsBasis.h
@@ -142,6 +142,14 @@ public:
         return result;
     }
 
+    /// Evaluate the third derivative of a single basis function \a i at points \a u.
+    gsMatrix<T> deriv3Single(index_t i, const gsMatrix<T> & u) const
+    {
+        gsMatrix<T> result;
+        this->deriv3Single_into(i, u, result);
+        return result;
+    }
+
     /** \brief Number of active basis functions at an arbitrary parameter value.
      *
      *  Usually, this is used for getting the active functions on one
@@ -663,10 +671,17 @@ public:
      */
     virtual void deriv2_into(const gsMatrix<T> & u, gsMatrix<T>& result ) const;
 
+    virtual void deriv3_into(const gsMatrix<T> & u, gsMatrix<T>& result ) const;
+
     /// @brief Evaluate the (partial) derivatives of the \a i-th basis function
     /// at points \a u into \a result.
     // hessianSingle_into
     virtual void deriv2Single_into(index_t i,
+                                   const gsMatrix<T> & u,
+                                   gsMatrix<T>& result ) const;
+
+    /// Like \a deriv2Single_into but for third derivatives.
+    virtual void deriv3Single_into(index_t i,
                                    const gsMatrix<T> & u,
                                    gsMatrix<T>& result ) const;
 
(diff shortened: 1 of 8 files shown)


src/gsNurbs/gsDeboor.hpp
```diff
@@ -218,6 +218,34 @@ void gsTensorDeriv2_into(const gsMatrix<T>& u,
                 result(k, j) += coefs(ind(i, j)) * deriv2(k + i * n2der, j);
 }
 
+// =============================================================================
```
filiatra, line 221: For the linear combination of basis values, please use the existing function:

gsBasis::linearCombination_into

and remove this one to avoid duplication

src/gsCore/gsBasis.h
```diff
@@ -142,6 +142,14 @@ class gsBasis : public gsFunctionSet<T>
         return result;
     }
 
+    /// Evaluate the third derivative of a single basis function \a i at points \a u.
```
filiatra, line 145: We may note in the comment that the single versions can be much slower than the all-at-once

src/gsTensor/gsTensorBasis.hpp
```diff
@@ -498,6 +498,96 @@ void gsTensorBasis<d,T>::deriv2Single_into(index_t i,
     }
 }
 
+template<short_t d, class T>
+void gsTensorBasis<d,T>::deriv3Single_into(index_t i,
+                                           const gsMatrix<T> & u,
+                                           gsMatrix<T>& result) const
+{
```
filiatra, line 505: The third derivatives are already computed by 

gsTensorBasis::evalAllDers(u,3,result)

The order for second derivatives is pure-first and then mixed.

From the 3rd derivative and upwards the derivatives are stored in lexicographic order.

The new functions should not follow the special rule of the deriv2 case, but they should return the derivatives in lex-order to comply with  evalAllDers.

unittests/gsDeriv3_test.cpp
```diff
@@ -0,0 +1,136 @@
+/** @file gsDeriv3_test.cpp
```
filiatra, line 1: In this unit test the result of the proposed deriv3 function should be check to be equal to the matrix all[3] that is returned provided by  

std::vector<gsMatrix> all =  gsTensorBasis::evalAllDers(u,3);

filiatra (changes requested):

dominik-mokris-mtu:
Following @filiatra's suggestion, we've modified our code to use `evalAllDers_into` instead of `deriv3_into`. As such, we don't need the PR, closing.

I will also delete the branch feature/deriv3 but the code could be revived from its final commit (b6fbf07e5d02c53769b3dd572e9514267826a656).

PR #893, opened 2026-08-27
Add custom quadrature factory to expression assembler

mhodzelmans:
NEW: the ability to inject a quadrature rule into the assembler. 

With the current assembler, quadrature rules can only be requested through an options list, from a predefined set within gsQuadrature. This means that any (new) rule must be incorporated in gsQuadrature, and therefore in the gismo core.

This commit allows quadrature rules to be defined outside the core.

### DIFF
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -23,9 +23,6 @@
 
 #include <gsMatrix/gsFiberMatrix.h>
 
-#include <functional>
-#include <utility>
-
 namespace gismo
 {
 
@@ -154,7 +151,7 @@ public:
     /// quadrature. The factory is invoked only when a new rule is needed, never
     /// in an element or quadrature-point loop.
     void setQuadratureFactory(QuadratureFactory factory)
-    { m_quadratureFactory = std::move(factory); }
+    { m_quadratureFactory = give(factory); }
 
     /// @brief Restores the standard gsQuadrature/options-based rules.
     void clearQuadratureFactory()
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -23,6 +23,9 @@
 
 #include <gsMatrix/gsFiberMatrix.h>
 
+#include <functional>
+#include <utility>
+
 namespace gismo
 {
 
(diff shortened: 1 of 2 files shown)


src/gsAssembler/gsExprAssembler.h
```diff
@@ -23,6 +23,9 @@
 
 #include <gsMatrix/gsFiberMatrix.h>
 
+#include <functional>
+#include <utility>
```
hverhelst: These might be included elsewhere probably, so we can remove them here.
Please test this

src/gsAssembler/gsExprAssembler.h
```diff
@@ -123,6 +148,22 @@ class gsExprAssembler
     /// Returns a reference to the options structure
     gsOptionList & options() {return m_options;}
 
+    /// @brief Installs a custom quadrature-rule factory.
+    ///
+    /// Passing an empty factory restores the standard option-driven
+    /// quadrature. The factory is invoked only when a new rule is needed, never
+    /// in an element or quadrature-point loop.
+    void setQuadratureFactory(QuadratureFactory factory)
+    { m_quadratureFactory = std::move(factory); }
```
hverhelst: Probably `give` works as well, which would be our preference.
```suggestion
    { m_quadratureFactory = give(factory); }
```

unittests/gsExprAssembler_test.cpp
```diff
@@ -13,9 +13,88 @@
...
+                return gsExprAssembler<real_t>::QuadratureRulePtr(
+                    new MidpointRule<real_t>(basis.dim()));
+            });
+
+        CHECK(custom.hasCustomQuadrature());
+        custom.assemble(uCustom * uCustom.tr());
+        const gsSparseMatrix<real_t> Mcustom = custom.matrix();
+
+        CHECK(calls.load() > 0);
+        CHECK(contextIsCorrect.load());
```
hverhelst, line 85: I was wondering whether these are really needed, but it might indeed be good to check whether the factory is actually called.

unittests/gsExprAssembler_test.cpp
```diff
@@ -13,9 +13,88 @@
...
+
+        custom.clearQuadratureFactory();
+        CHECK(!custom.hasCustomQuadrature());
+        custom.clearMatrix();
+        custom.assemble(uCustom * uCustom.tr());
+        const gsSparseMatrix<real_t> Mrestored = custom.matrix();
+
+        CHECK_EQUAL(Mstandard.rows(), Mrestored.rows());
+        CHECK_EQUAL(Mstandard.cols(), Mrestored.cols());
+        CHECK((Mstandard - Mrestored).norm() < 1e-14);
```
hverhelst, line 97: Here we could also assert that the number of calls should be the same as before clearing it. Then we are sure that it is not called anymore.

src/gsAssembler/gsExprAssembler.h
mhodzelmans: Done. The new test doesn't need it

unittests/gsExprAssembler_test.cpp
mhodzelmans: If the matrices match that doesn't guarantee that the new rule is actually being used, so I do think it's valuable

hverhelst (approved):

PR #895, opened 2026-09-04
Hotfixes: numeric limits and space sizes

hverhelst:
This PR fixes some issues with numeric limits and block sizes in `gsExprAssembler`, and the `max_op` in `gsExprEvaluator`.

----

NEW: 

IMPROVED: 

FIXED: 
For multiple spaces (e.g., vector space v and scalar space q) `_blockDims` and `resetDimensions` would use the wrong sizes for rows and columns. They'd use `dim*freeSize` but `freeSize` does already include `dim`. The `numDofs` in `gsExprAssembler` already does this correct.
In `gsExprEvaluator` the `max_op` would not work on pure negative functions. It was initialized with `numeric_limits::min` which is `1e-300` which can still be the maximum, for any function with a maximum <0 in the domain. It is replaced with `numeric_limits::lowest` which is minus infinity. 
 

API:

----

request:
-----

### DIFF
--- a/src/gsAssembler/gsExprAssembler.h
+++ b/src/gsAssembler/gsExprAssembler.h
@@ -543,10 +543,10 @@ private:
         {
             rowSizes.resize(m_vrow.size());
             for (index_t r = 0; r != rowSizes.size(); ++r) // for all row-blocks
-                rowSizes[r] = m_vrow[r]->dim() * m_vrow[r]->mapper.freeSize();
+                rowSizes[r] = m_vrow[r]->mapper.freeSize();
             colSizes.resize(m_vcol.size());
             for (index_t c = 0; c != colSizes.size(); ++c) // for all col-blocks
-                colSizes[c] = m_vcol[c]->dim() * m_vcol[c]->mapper.freeSize();
+                colSizes[c] = m_vcol[c]->mapper.freeSize();
         }
     }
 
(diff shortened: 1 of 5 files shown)


filiatra (approved):

src/gsAssembler/gsExprAssembler.h
```diff
@@ -418,6 +417,14 @@ class gsExprAssembler
...
             }
         }
+
+        // Both paths leave m_fmatrix inconsistent with the cached m_matrix --
+        // one zeroes the values, the other resizes and drops the pattern -- so
+        // the cache is invalidated here rather than per branch. matrix() is
+        // `m_modified ? makeMatrix() : m_matrix`, so a path that resizes
+        // without setting this returns the stale m_matrix, still 0x0 on the
+        // first initSystem() since nothing has assembled into it yet.
+        m_modified = true;
```
hverhelst, line 421: paths leave m_fmatrix inconsistent with the cached m_matrix -- zeroes the values, the other resizes and drops the pattern -- so cache is invalidated here rather than per branch. matrix() is m_modified ?  makeMatrix() : m_matrix`, so a path that resizes setting this returns the stale m_matrix, still 0x0 on the initSystem() since nothing has assembled into it yet.

filiatra (approved):

PR #900, opened 2026-09-14
Bugfix in `gsDirichletValuesByTPInterpolation`

hverhelst:
The `gsDirichletValuesByTPInterpolation` is based on `TPInterpolation`, indeed, tensor-product interpolation. Since it calls
```
for ( int i=0; i < parDim; ++i)
{
    if ( i==dir )
    {
        b[0] = ( basis.component(i).support() ) (0, param);
        rr.push_back(b);
    }
    else
    {
        rr.push_back( basis.component(i).anchors().transpose() );
    }
}
```
It takes `basis.component(i)` for `gsHTensorBasis` which is
```
m_bases[ this->maxLevel() ]->component(i);
```
So it interpolates on the wrong anchors.

In addition, the `fixedDofs.at(ii) = dVals.at(l);` was wrong, as it would only touch column 0. This is fixed now by changing to
```
fixedDofs.at(ii) = dVals(l, cc);
```

Added tests that will fail without these changes.

----

NEW:

IMPROVED:

FIXED: 
Dirichlet values by interpolation for HSplines.

API:

---

**Please consider the following checklist before issuing a pull
request:**
- [X] Have you added an explanation of what your changes do and why
  you'd like us to include them?
- [X] Have you documented any new codes using Doxygen comments?
- [X] Have you written new tests or examples for your changes?
-----

PR #901, opened 2026-09-18
HotFix add error for gsHTensorBasis in gsDirichletValues

hverhelst:
Fix to warn for the behavior in #900 

NEW:

IMPROVED: 
Add error for `gsHTensorBasis` in `gsDirichletValues`

FIXED: 

API:


filiatra (changes requested):
In file gsTensorBasis.h line 55
```
   domainIter beginBdr(const boxSide bs) const override
    { return domainIter(new gsTensorDomainBoundaryIterator<T,D,knotIter>(*this, bs)); ```
move the body of this function to the .hpp file to avoid errors when including the header

filiatra (approved):

